PR #2, opened 2024-02-13
feat(PE-5652): scaffold initial providers

atticusofsparta:
Scaffolds the initial providers
- ArIo - main provider
- ArNSRemoteCache - remote evaluation interface for the arns-service

getContractState currently the only method implemented on IContractStateProvider which the two above providers implement.


dtfiedler (approved):
off we go!

PR #57, opened 2024-04-25
fix(utils): add writeInteraction types and update base64url logic, disable bunding on all writeInteractions

dtfiedler:


src/common/contracts/warp-contract.ts
```diff
@@ -89,12 +87,14 @@ export class WarpContract<T>
     }
     const warpSigner = new Signature(this.warp, {
       signer: async (tx: Transaction) => {
+        const publicKey = await signer.publicKey;
```
atticusofsparta, line 90: do we need to call set public key for arconnect signer?

atticusofsparta (approved):

src/common/contracts/warp-contract.ts
dtfiedler: https://github.com/ar-io/ar-io-sdk/pull/57/files#diff-6322e9c5f09198a48852508064d79447ceeb2f052204577f63789722aff64a3bL88

atticusofsparta (dismissed):

PR #72, opened 2024-05-02
fix(base64): use arweave utils for browser friendly buffer conversion

atticusofsparta:


fedellen (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ar.io/sdk",
-  "version": "1.0.4",
+  "version": "1.0.3",
```
fedellen, line 3: should this be going to alpha?

src/utils/base64.ts
```diff
@@ -14,14 +14,16 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+import Arweave from 'arweave';
```
fedellen, line 17: this import can always be tricky in a non web/node directory

atticusofsparta: gotchya - still figuring out if this is the best way to do it, considering other options like https://www.npmjs.com/package/base64url at the moment.

dtfiedler:
Replace with https://github.com/ar-io/ar-io-sdk/pull/73

PR #78, opened 2024-05-03
fix(joinNetwork): make params required - properties and note

atticusofsparta:


src/common/ar-io.ts
```diff
@@ -634,12 +634,36 @@ export class ArIOWritable extends ArIOReadable implements ArIOWriteContract {
     * arIO.joinNetwork(jointNetworkParams);
    * ```
    */
-  async joinNetwork(
-    params: JoinNetworkParams,
-  ): Promise<WriteInteractionResult> {
+  async joinNetwork({
+    qty,
+    allowDelegatedStaking,
```
atticusofsparta, line 639: moved to verbose params here because the types were not visible.

src/common.ts
```diff
@@ -222,9 +224,13 @@ export interface ArIOWriteContract {
 
 export type WriteInteractionResult = Transaction | DataItem;
 
-export type JoinNetworkParams = GatewayConnectionSettings &
-  GatewayStakingSettings &
-  GatewayMetadata & { qty: number };
+// Utility type for dev experience: shows all fields of a type rather than the composed type names
+export type VisibleTypes<T> = { [P in keyof T]: T[P] };
```
atticusofsparta, line 228: This consumes an intersecting type and makes it visible to intellisense - it should rebuild the exact type.

src/contract-state.ts
```diff
@@ -69,9 +69,8 @@ export type GatewayStakingSettings = {
 
 export type GatewayMetadata = {
   label: string;
-  note?: string;
-  properties?: string;
-  observerWallet?: WalletAddress;
+  note: string;
+  properties: string;
```
atticusofsparta, line 73: These properties are required to join the network and are non nullable in gateway settings updates

src/common.ts
```diff
@@ -238,11 +244,14 @@ export type UpdateGatewaySettingsParamsBase = {
   properties?: string;
   protocol?: AllowedProtocols;
   autoStake?: boolean;
+  observerWallet?: WalletAddress;
 };
 
 // Utility type to require at least one of the fields
-export type AtLeastOne<T, U = { [K in keyof T]-?: T[K] }> = Partial<U> &
-  { [K in keyof U]: Required<Pick<U, K>> }[keyof U];
+export type AtLeastOne<
```
atticusofsparta, line 251: I HAD checked this before but not down enough avenues - this now properly accepts multiple properties as well as a single property, requiring them to be non nullable, and throws on empty objects.

src/common.ts
dtfiedler: I don't love these bc they are 1. hard to read/understand and bc of that 2. hard to change with confidence

I'll offer a commit with what I think might align with what we are trying to acheive.

atticusofsparta: I get what you mean, and im not a fan either - this type serves . Not trying to be pedantic but does this improve readability at all? 
```typescript
export type ShowGranularPropertyTypes<ObjectToShowGranularly> = { [PropertyToShow in keyof ObjectToShowGranularly]: ObjectToShowGranularly[PropertyToShow] };
```

PR #86, opened 2024-05-21
fix(types): add delegated gateway type

dtfiedler:


fedellen (approved):

src/contract-state.ts
```diff
@@ -60,6 +61,12 @@ export type Gateway = {
   weights: ObserverWeights;
 };
 
+export type GatewayDelegate = {
```
kunstmusik, line 64: Could you also modify Gateway to have:

```
delegtes: Record<string, GatewayDelegate>
```

dtfiedler: ah, good catch

dtfiedler: [ac7e924](https://github.com/ar-io/ar-io-sdk/pull/86/commits/ac7e924f24ba6133b2917de8416c37bf6560cdd4)

PR #87, opened 2024-05-21
feat(PE-6063): add extendLease and increaseUndernameSupport apis

atticusofsparta:


dtfiedler:
let's continue to call this a `fix` as it was missed in earlier releases and does not impact any existing APIs

dtfiedler (approved):

PR #99, opened 2024-05-24
chore: release v1.1.0

dtfiedler:
Details:
- supports tags in write interactions
- adds retries to HTTP requests
- adds util functions for smartweave (not necessary, but useful)
- syncState before read interactions (should alleviate evaluation issues when just doing read interactions)


dtfiedler:
@atticusofsparta once you confirm and validate new APIs we can release

atticusofsparta:
shouldnt this be 1.0.9?

PR #104, opened 2024-05-24
feat(PE-6111): get contracts for wallet like WarpContract

atticusofsparta:


src/common.ts
```diff
@@ -84,6 +90,8 @@ export interface BaseContract<T> {
 }
 
 export interface ReadContract {
+  // Expectation of this method is to return a list of the contracts owned by the user on the specific protocol (eg smartweave)
+  getContracts(params: { address: string }): Promise<DataProtocolTransaction[]>;
```
dtfiedler, line 94: i don't think this should exist on ReadContract

src/utils/graphql/common.ts
```diff
@@ -0,0 +1,141 @@
...
+    cursor =
+      response.data.data.transactions.edges[MAX_REQUEST_SIZE - 1]?.cursor;
+    hasNextPage =
+      response.data.data.transactions.pageInfo?.hasNextPage ?? false;
+  } while (hasNextPage);
+
+  return [...protocolTxs];
+}
+
+export async function getSmartweaveContractsFromGQL(params: {
```
dtfiedler, line 132: let's just put this all into one function, i don't think we need to separate them. there is a lot we are assuming in the `getDataProtocolTransactionsByOwner`, e.g. `bundledIn: null` so it's not really giving us any abstraction

src/common/error.ts
```diff
@@ -50,3 +50,5 @@ export class InvalidContractConfigurationError extends BaseError {
     super(INVALID_CONTRACT_CONFIGURATION_ERROR);
   }
 }
+
+export class EvaluationTimeoutError extends BaseError {}
```
dtfiedler, line 54: where is this being thrown?

src/common.ts
```diff
@@ -44,6 +44,12 @@ export type SortKey = string;
 export type WalletAddress = string;
 export type TransactionId = string;
 
+export type DataProtocolTransaction = {
+  id: string;
+  tags: { name: string; value: string }[];
+  data: number;
+};
```
dtfiedler, line 51: does arbundles or arweave-js export a type like this? ideally the SDK doesn't have to create it's own types for common ones

src/common/error.ts
atticusofsparta: in the query function

dtfiedler: It's not actually an evaluation issue, it's an aborted request. So we should throw an AbortError

src/common.ts
atticusofsparta: Yes - updated.

Uses Pick with GQLNodeInterface.

src/utils/graphql/common.ts
```diff
@@ -0,0 +1,124 @@
...
+      {
+        signal,
+      },
+    );
+
+    if (status == 429) {
+      // Rate limited, wait for 10 seconds and retry
+      await new Promise((resolve) => setTimeout(resolve, 10_000));
+      continue;
+    }
```
dtfiedler, line 97: it would be nice if you could provide a retry strategy via the params, not necessary for now, but can add as a TODO

atticusofsparta: sure - will add the todo. In the future what I would like to see is a specific gql provider that consumes queryBuilder() and has that retry stuff internally.

dtfiedler (changes requested):
some comments:
1. throw an AbortError on aborted request and not `EvalutionTimeoutIssue`
2. move `MAX_REQUEST_SIZE` to smartweave.ts
3. remove comments

src/utils/graphql/smartweave.ts
```diff
@@ -0,0 +1,389 @@
...
+ * @param address - address of the wallet to get transactions for
+ * @param contractTxId - contract tx id to filter the transactions
+ * @param signal - signal to abort the request
+ * @returns @type {Promise<DataProtocolTransaction[]>} - returns the list of data protocol txs, for e.g smartweave contracts
+ */
+export async function getSmartweaveTransactionsFromGQL({
+  address,
+  contractTxId,
+  arweave,
+  signal,
```
dtfiedler, line 136: we may want to hard code this to be like 10 seconds or something (e.g `signal = AbortSignal.timeout(10_000)`)

atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/104/commits/6ef3aa5b3553ed4c67524f56a6ff223fd60dc0a8

dtfiedler (commented):
can you add an integration test for this? you should be able to use the existing deployed contracts

dtfiedler (approved):

PR #108, opened 2024-05-29
feat(ao): io/ao contract classes

dtfiedler:
A WIP of accommodating the new AO/IO contract.

Aiming to keep the interfaces very similar, minus a few response types changing so there is minimal lift from downstream clients.

Tickets: 
- https://ardrive.atlassian.net/browse/PE-6242
- https://ardrive.atlassian.net/browse/PE-6093

Related PRs:
- Network Portal - https://github.com/ar-io/network-portal/pull/11
- Observer - https://github.com/ar-io/ar-io-observer/pull/40


src/common/contracts/ao-contract.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import { dryrun } from '@permaweb/aoconnect';
```
atticusofsparta, line 17: I recommend a pattern of something like this that can consume connection props and provide access to the API's provided by ao connect: https://github.com/project-kardeshev/ao-rtc/blob/main/src/ao-provider.ts

src/common/contracts/ao-contract.ts
```diff
@@ -0,0 +1,77 @@
...
+    logger = new DefaultLogger(),
+  }: {
+    processId: string;
+    logger?: DefaultLogger;
+  }) {
+    this.processId = processId;
+    this.logger = logger;
+  }
+
+  async getState(): Promise<T> {
```
atticusofsparta, line 37: Just a note on this, I think we are currently limited to 10mb returns here.

dtfiedler: [05b07cf](https://github.com/ar-io/ar-io-sdk/pull/108/commits/05b07cfbb1d974e708108c8239d8214d6c50b418)

src/common/contracts/ao-contract.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const data: K = JSON.parse(result.Messages[0].Data);
+    return data;
+  }
+
+  async send<K>({
+    tags,
+    data,
+    signer,
+  }: {
```
atticusofsparta, line 104: wondering if we should add a `withResult` flag here that takes the msg ID and awaits the evaluation, returning the result of the interaction. It is something we can do manually but may be useful for control flow purposes.

src/constants.ts
```diff
@@ -31,4 +31,6 @@ export const ARNS_TESTNET_REGISTRY_TX =
 export const ARNS_DEVNET_REGISTRY_TX =
   '_NctcA2sRy1-J4OmIQZbYFPM17piNcbdBPH2ncX2RL8';
 
+export const ioDevnetProcessId = 'GaQrvEMKBpkjofgnBi_B3IgIDmY_XYelVLB6GcRGrHc';
```
kunstmusik, line 34:
```suggestion
export const IO_DEVNET_PROCESS_ID = 'GaQrvEMKBpkjofgnBi_B3IgIDmY_XYelVLB6GcRGrHc';
```

PR #111, opened 2024-05-30
feat(write): add tags support to write interactions on warp-contract and saveObservations

atticusofsparta:


dtfiedler:
this looks good so far, you'll need to allow `tags` on the API write interactions as well

src/common.ts
```diff
@@ -81,10 +81,19 @@ export type ReadParameters<Input> = {
...
+export type WriteOptions = {
+  tags?: { name: string; value: string }[];
+};
+
 export type WriteParameters<Input> = WithSigner<
   Required<ReadParameters<Input>>
 >;
 
+export type WriteArguments<T> = [
+  EvaluationParameters<WriteParameters<T>>,
```
dtfiedler, line 93: I'm just realizing this now, but Write interactions should not support evaluation options

atticusofsparta: You are correct, fixing.

README.md
```diff
@@ -212,6 +212,34 @@ const ioValue = new mIOToken(mIOValue).toIO();
...
+});
+const res = await ant.setName(
+  {
+    name: 'My awesome new name',
+  },
+  { tags: [{ name: 'App-Name', value: 'My-Awesome-App' }] },
+);
+```
+
+</details>
```
dtfiedler, line 241: I'm not sure we need an entire section dedicated to tags. Let's just add a note to all the APIs that do support them and add to their examples
```
// optional additional tags
  { tags: [{ name: 'App-Name', value: 'My-Awesome-App' }] },
```
```

File: src/common.ts [hunk 3]
```diff
@@ -98,13 +104,9 @@ export interface ReadContract {
...
 export interface WriteContract {
-  writeInteraction<Input>({
-    functionName,
-    inputs,
-    evaluationOptions,
-  }: EvaluationParameters<
-    WriteParameters<Input>
-  >): Promise<WriteInteractionResult>;
+  writeInteraction<Input>(
+    ...[{ functionName, inputs }, options]: WriteArguments<Input>
```
dtfiedler, line 108: using a spread operator here is something i haven't seen before. can you confirm this is equivalent to `writeInteraction<Input>({ functionName, inputs}, options)`?

dtfiedler (commented):
the implemenation looks good, small comment on the README and a question r.e. the type created - otherwise looks good. can we add tests that validate a write interaction with tags by checking them in GQL?

src/common.ts
atticusofsparta: eh kinda, it actually looks like `0_0_WriteParameters<Input>, 0_1_WriteOptions` - it is technically the correct type and is human readable, just a bit ugly. I mainly did it this way instead of `(params: WriteArguments<Input>)` to enable verbosity for the positional parameters in the type def.

I tried it without the spread operator as well but typescript then considers it a positional argument at the zeroth index.

dtfiedler: hm, i think i'd prefer `writeInteraction<Input>({ functionName, inputs}, { tags })`

tests/integration/ar-io-writable.test.ts
```diff
@@ -132,12 +136,34 @@ describe('ArIOWriteable', () => {
...
+        tags: [
+          {
+            name: 'ar-io-sdk-write',
+            value: AR_IO_CONTRACT_FUNCTIONS.INCREASE_UNDERNAME_LIMIT,
+          },
+        ],
+      },
+    );
+
+    const tx = await arweave.transactions.get(await res.id);
```
dtfiedler, line 154:
```suggestion
    const tx = await arweave.transactions.get(res.id);
```

atticusofsparta: for some reason res.id can be a promise or a string

atticusofsparta: Yeah because we return Transaction OR DataItem, we are chained to this type 
image
I might be able to add an overload to fix this

dtfiedler (approved):

PR #122, opened 2024-06-18
feat(PE-6240): add spawn ant util

atticusofsparta:


src/utils/ao.ts
```diff
@@ -0,0 +1,151 @@
...
+
+      if (evalResult instanceof Error && retries < maxRetries) {
+        options.logger?.error(
+          `Error messaging process: ${evalResult.message}. Retrying in 2 seconds.`,
+          {
+            processId,
+            error: evalResult,
+          },
+        );
+        await new Promise((resolve) => setTimeout(resolve, retryDelay));
```
dtfiedler, line 109: use exponential backoff - https://github.com/ar-io/ar-io-sdk/blob/alpha/src/common/contracts/ao-process.ts#L134-L136

src/utils/ao.ts
```diff
@@ -0,0 +1,151 @@
...
+        { name: 'App-Name', value: 'ArNS-AO-ANT' },
+      ],
+    });
+
+    let antCodeLoaded = false;
+    let retries = 0;
+    const retryDelay = 5000;
+    const maxRetries = 5;
+    // retry is necessary since the process may not be ready to receive messages immediately
+    while (antCodeLoaded === false) {
```
dtfiedler, line 90: i would do something more like this try logic - https://github.com/ar-io/ar-io-sdk/blob/alpha/src/common/contracts/ao-process.ts#L97-L140

src/ant-lua-code.ts
dtfiedler, line 1: don't hardcode this here, just hardcode the current tx id of the source code on arweave and fetch it from arweave when spawning the ant

src/utils/ao.ts
```diff
@@ -0,0 +1,151 @@
...
+ * ```
+ */
+export async function spawnANT(
+  {
+    module = AOS_MODULE_ID,
+    luaString = ANT_LUA,
+    ao = connect(),
+    scheduler = DEFAULT_SCHEDULER_ID,
+    signer,
+    state,
```
dtfiedler, line 46: add the txId of the source code for the bundled lua code and default it to our current tx. then store it locally on the filesystem in `tmp/arns/ant.lua` file

src/utils/ao.ts
atticusofsparta: moved the logic for this to inside AOProcess.send (retries/exponential backoff)

src/utils/ao.ts
atticusofsparta: added a todo for this.

src/utils/ao.ts
```diff
@@ -0,0 +1,95 @@
...
+  DEFAULT_SCHEDULER_ID,
+} from '../constants.js';
+import { ANTState } from '../contract-state.js';
+import { ContractSigner } from '../types.js';
+
+/**
+ *
+ * @param param0 spawn and message args
+ * @param connectionConfig
+ * @returns @type {Promise<string>} processId
```
dtfiedler, line 31: nit: remove or update

src/utils/ao.ts
```diff
@@ -0,0 +1,95 @@
...
+      { name: 'App-Name', value: 'ArNS-ANT' },
+      { name: 'Source-Code-TX-ID', value: luaCodeTxId },
+    ],
+    data: luaString,
+    signer,
+  });
+
+  if (state) {
+    await aosClient.send({
+      tags: [{ name: 'Action', value: 'Initialize-State' }],
```
dtfiedler, line 88: wondering if we should add `Initialized-For` tag that is the address of the signer provided

src/utils/ao.ts
```diff
@@ -0,0 +1,95 @@
...
+  {
+    module = AOS_MODULE_ID,
+    luaCodeTxId = ANT_LUA_ID,
+    scheduler = DEFAULT_SCHEDULER_ID,
+    signer,
+    state,
+  }: {
+    module: string;
+    luaCodeTxId: string;
+    aoClient: AOProcess;
```
dtfiedler, line 43:
```suggestion
```

src/utils/ao.ts
```diff
@@ -0,0 +1,95 @@
...
+    scheduler: string;
+    signer: ContractSigner;
+    state?: ANTState;
+  },
+  connectionConfig?: {
+    CU_URL: string;
+    MU_URL: string;
+    GATEWAY_URL: string;
+    GRAPHQL_URL: string;
+  },
```
dtfiedler, line 53:
```suggestion
export async function spawnANT(
  {
    moduleId = AOS_MODULE_ID,
    luaCodeTxId = ANT_LUA_ID,
    scheduler = DEFAULT_SCHEDULER_ID,
    ao = connect(),
    signer,
    state,
  }: {
    moduleId: string;
    luaCodeTxId: string;
    ao: AOClient;
    scheduler: string;
    signer: ContractSigner;
    state?: ANTState;
  },
```

and remove the subbing of AOProcess and `spawn` using `ao.connect`

src/utils/ao.ts
atticusofsparta: Good idea, but I would say rather that it should be the contract ID or process ID that the state came from

src/common/contracts/ao-process.ts
```diff
@@ -17,50 +17,32 @@
...
-      CU_URL: connectionConfig?.CU_URL,
-      GATEWAY_URL: connectionConfig?.GATEWAY_URL,
-      GRAPHQL_URL: connectionConfig?.GRAPHQL_URL,
-    });
+    this.ao = ao;
   }
 
   // TODO: could abstract into our own interface that constructs different signers
-  async createAoSigner(
+  static async createAoSigner(
```
dtfiedler, line 45: why does this need to be static?

src/utils/ao.ts
dtfiedler: that's a good point, thinking we should optionally allow `contractTxId` and add it as tag

src/common/contracts/ao-process.ts
atticusofsparta: When I was writing the spawnANT util, i needed to create a signer. I think we should make this static or a util - it doesnt actually use any of the class properties.

A WalletFactory class might be a good idea, so we can do like AOSignerFactory.arweave(ContractSigner) and AOSignerFactory.ethereum(ContractSigner)

src/utils/ao.ts
atticusofsparta: I went with "stateContractTxId" to avoid confusion with `processId`.

src/common/contracts/ao-process.ts
dtfiedler: we can address in a separate PR

dtfiedler (approved):

PR #132, opened 2024-06-25
fix(ao ant): add handler for get state

atticusofsparta:


src/common/ant-ao.ts
```diff
@@ -46,6 +46,14 @@ export class AoANTReadable implements AoANTRead {
     }
   }
 
+  async getState(): Promise<ANTState> {
+    const tags = [{ name: 'Action', value: 'State' }];
+    const res = await this.process.read<ANTState>({
```
dtfiedler, line 51: this type is not the same as ANTState - https://github.com/ar-io/ar-io-ant-process/blob/main/src/common/main.lua#L404-L415

dtfiedler (approved):

PR #135, opened 2024-06-26
fix(ao reads): safely parse json

atticusofsparta:


src/common/contracts/ao-process.ts
```diff
@@ -102,10 +103,10 @@ export class AOProcess implements AOContract {
         }
 
         this.logger.debug(`Read interaction result`, {
-          result: result.Messages[0]?.Data,
+          result: result.Messages[0].Data,
         });
 
-        const response: K = JSON.parse(result.Messages[0]?.Data);
+        const response: K = safeDecode<K>(result.Messages[0].Data);
```
dtfiedler, line 109: what if `K` is not provided?

atticusofsparta: Good point, set it as unknown as default generic in util

dtfiedler (approved):

PR #136, opened 2024-06-26
fix(ant): add event emitter util for fetching ants

dtfiedler:


src/utils/graphql/processes.ts
```diff
@@ -42,19 +43,78 @@ export const getANTProcessesOwnedByWallet = async ({
...
+    uniqueContractProcessIds.map(async (processId) =>
       throttle(async () => {
         const ant = ANT.init({
           processId,
         });
-        const owner = await ant.getOwner();
-        const controllers = await ant.getControllers();
-        return owner === address || controllers.includes(address);
+        const [owner, controllers = []] = await Promise.all([
+          ant.getOwner().catch(() => undefined),
```
atticusofsparta, line 53: can use get state instead to perform 1 query instead of 2

atticusofsparta (approved):

PR #143, opened 2024-07-01
fix(io): add epoch-settings api and tests

dtfiedler:


src/common/io.ts
```diff
@@ -124,6 +125,37 @@ export class IOReadable implements AoIORead {
       tags: [{ name: 'Action', value: 'Info' }],
     });
   }
+
+  async getEpochSettings(params?: EpochInput): Promise<AoEpochSettings> {
+    const allTags = [
+      { name: 'Action', value: 'Epoch-Settings' },
+      {
+        name: 'Timestamp',
```
atticusofsparta, line 133: Have you tested this live? "Timestamp" is an overlapping tag I think.

src/common/io.ts
```diff
@@ -124,6 +125,37 @@ export class IOReadable implements AoIORead {
...
+    ];
+
+    const prunedTags: { name: string; value: string }[] = allTags.filter(
+      (tag: {
+        name: string;
+        value: string | undefined;
+      }): tag is { name: string; value: string } => tag.value !== undefined,
+    );
+
+    return this.process.read<AoEpochSettings>({
```
atticusofsparta, line 155: Wondering if process.read should auto-prune tags, think we are repeating this frequently yeah?

src/common/io.ts
dtfiedler: Timestamp is not overlapping on reads

PR #147, opened 2024-07-02
fix(imports): update named web exports to properly provide esm file i…

dtfiedler:
Note: this is a breaking change


dtfiedler:
will be merged in #152

PR #154, opened 2024-07-08
feat(PE-6314): add evolve util

atticusofsparta:


dtfiedler:
Are the changes in [this PR](https://github.com/ar-io/ar-io-ant-process/pull/5) required for this to work? have you tested against existing ANTs?

src/utils/ao.ts
```diff
@@ -85,3 +85,51 @@ export async function spawnANT({
...
+  signer: ContractSigner;
+  processId: string;
+  luaCodeTxId?: string;
+  ao?: AoClient;
+}): Promise<string> {
+  //TODO: cache locally and only fetch if not cached
+  const luaString = (await defaultArweave.transactions.getData(luaCodeTxId, {
+    decode: true,
+    string: true,
+  })) as string;
```
dtfiedler, line 104: does `arweave` support generics on `getData`?

src/utils/ao.ts
```diff
@@ -85,3 +85,51 @@ export async function spawnANT({
...
+
+  const { id } = await aosClient.send({
+    tags: [
+      { name: 'Action', value: 'Eval' },
+      { name: 'App-Name', value: 'ArNS-ANT' },
+      { name: 'Source-Code-TX-ID', value: luaCodeTxId },
+    ],
+    data: luaString,
+    signer,
+  });
```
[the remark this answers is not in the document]
atticusofsparta: It occured to me that would prevent writing twice, but in the event that it does not have the id you wind up actually sending three network calls instead of two, so I figured it was more efficient to do it this way... now its less network calls but the second call is heavier because its sending the code twice. Gas is also something to think about here, since that info check costs gas, and im not sure what the ratio is compared to sending eval twice, or verifying with info first.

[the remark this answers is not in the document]
dtfiedler: i favor calling getInfo to avoid unnecessary evals

examples/cjs/index.cjs
```diff
@@ -20,8 +29,28 @@ const { IO, ioDevnetProcessId, Logger } = require('@ar.io/sdk');
...
+
+  const processId = await spawnANT({
+    signer: new ArweaveSigner(jwk),
+  });
+
+  await evolveANT({
+    processId,
+    signer: new ArweaveSigner(jwk),
+    luaCodeTxId: '40ehDFzeiOjirz2V7R7mn33D3W97-j_nOJ3pc4b2JXs',
+  });
```
dtfiedler, line 48: i don't think we need to include this here

examples/cjs/index.cjs
```diff
@@ -25,11 +33,23 @@ const { IO, ioDevnetProcessId, Logger } = require('@ar.io/sdk');
...
+    host: 'arweave.net',
+    port: 443,
+    protocol: 'https',
+  });
+
+  const jwk = await arweave.wallets.generate();
+
+  const processId = await spawnANT({
+    signer: new ArweaveSigner(jwk),
+  });
```
dtfiedler, line 46: i think we can remove this from the example as well and keep them scoped to read only APIs

dtfiedler: @atticusofsparta please remove

examples/cjs/index.cjs
```diff
@@ -25,11 +33,16 @@ const { IO, ioDevnetProcessId, Logger } = require('@ar.io/sdk');
   const observation = await arIO.getObservations({ epochIndex: 0 });
   const distributions = await arIO.getDistributions();
 
+  const arweave = Arweave.init({
+    host: 'arweave.net',
+    port: 443,
+    protocol: 'https',
+  });
```
dtfiedler, line 40:
```suggestion
```

dtfiedler (approved):

dtfiedler (approved):

dtfiedler (approved):

PR #160, opened 2024-07-09
fix: restore arweave path

kunstmusik:


tsconfig.web.json
```diff
@@ -6,6 +6,7 @@
     "outDir": "./lib/web",
     "moduleResolution": "bundler",
     "paths": {
+      "arweave": ["./node_modules/arweave/web/index.js"],
```
atticusofsparta, line 9: why?

dtfiedler: we produce CJS builds which don't allow for named exports to be used in the SDK - this ensures only the web exports from arweave are available and node ones are not for web builds

dtfiedler: it _shouldn't_ be an issue to not have it though - since we don't use any web specific implementations of `arweave` in our SDK (e.g. for signing) - and my hope was that we could remove it from this path resolution

PR #168, opened 2024-07-12
fix(arns): update event emitter to provide more events and logs while…

dtfiedler:
… loading arns records

Also allows clients to provide their own arns event emitter to handle loaded records

### DIFF
--- a/src/utils/processes.ts
+++ b/src/utils/processes.ts
@@ -19,7 +19,7 @@ import { pLimit } from 'plimit-lit';
 
 import { ANT } from '../common/ant.js';
 import { IO } from '../common/io.js';
-import { ILogger } from '../common/logger.js';
+import { ILogger, Logger } from '../common/logger.js';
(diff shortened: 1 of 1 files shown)


atticusofsparta (approved):

src/utils/processes.ts
```diff
@@ -203,6 +217,9 @@ export const fetchAllArNSRecords = async ({
           message: e?.message,
           stack: e?.stack,
         });
+
+        emitter?.emit('arnsError', `Error getting ArNS records: ${e}`);
```
atticusofsparta, line 221: maybe we can use the convention of "arns:<event>" here? I have seen it used elsewhere to seperate name spaces.

dtfiedler: [1d67dfe](https://github.com/ar-io/ar-io-sdk/pull/168/commits/1d67dfe1f7a4766bbce6cf5a55438f33423f660f)

atticusofsparta (approved):

PR #175, opened 2024-07-25
feat(signing): add window arweave wallet to available signing options

atticusofsparta:


src/token.ts
```diff
@@ -153,3 +153,10 @@ export class mIOToken extends PositiveFiniteInteger {
     return new IOToken(this.valueOf() / MIO_PER_IO);
   }
 }
+
+export type AoSigningFuction = (args: {
```
kunstmusik, line 157: Perhaps this should be called AoSigner, considering it's returned from createAoSigner()

src/common/ant.ts
```diff
@@ -227,14 +227,14 @@ export class AoANTReadable implements AoANTRead {
 }
 
 export class AoANTWriteable extends AoANTReadable implements AoANTWrite {
-  private signer: ContractSigner;
+  private signer: AoSigningFuction;
```
kunstmusik, line 230: might be good to rename "signer" to "aoSigner" through the SDK to make it self-documenting

examples/esm/index.mjs
```diff
@@ -66,10 +68,14 @@ import Arweave from 'arweave';
...
 
   const jwk = await arweave.wallets.generate();
-
-  const processId = await spawnANT({
-    signer: new ArweaveSigner(jwk),
-  });
+  let processId;
+  try {
+    processId = await spawnANT({
+      signer: AOProcess.createAoSigner(new ArweaveSigner(jwk)),
```
dtfiedler, line 74: i'm in favor of moving `createAoSigner` to a standalone utility function that we export and use in the class. there's no need for it to be static anymore

dtfiedler (changes requested):
see comment r.e. moving `createAoSigner` to a utility function

src/utils/ao.ts
```diff
@@ -120,3 +121,35 @@ export async function evolveANT({
...
 }
+
+export function createAoSigner(
+  signer: ContractSigner,
+): (args: {
+  data: string | Buffer;
+  tags?: { name: string; value: string }[];
+  target?: string;
+  anchor?: string;
+}) => Promise<{ id: string; raw: ArrayBuffer }> {
```
dtfiedler, line 132: update the type here now that we have `AoSigner`

examples/esm/index.mjs
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/175/commits/8e95eddf5723b0fdbf870688556f76a3e80bb6fa

src/utils/ao.ts
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/175/commits/8e95eddf5723b0fdbf870688556f76a3e80bb6fa

package.json
```diff
@@ -118,8 +118,9 @@
     "vite-plugin-node-polyfills": "^0.22.0"
   },
   "dependencies": {
-    "@permaweb/aoconnect": "^0.0.55",
+    "@permaweb/aoconnect": "^0.0.57",
     "arbundles": "0.11.0",
+    "arconnect": "^1.0.3",
```
dtfiedler, line 123: is this being used? if not required, remove from the build or put as a dev dep

atticusofsparta: This is the types package for arconnect, we do use it - will try as a dev dep

src/common/ant.ts
dtfiedler: i prefer `signer` to keep it abstracted from it's type, but a nit

package.json
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/175/commits/34f07d25024c080f529218569cba57041e34baa2

dtfiedler: should be fine as a dev dep if we are not explicitly exporting any arconnect types from our own build

src/common/io.ts
```diff
@@ -578,7 +579,7 @@ export class IOReadable implements AoIORead {
 
 export class IOWriteable extends IOReadable implements AoIOWrite {
   protected declare process: AOProcess;
-  private signer: ContractSigner;
+  private aoSigner: AoSigner;
```
dtfiedler, line 582: going to ask to revert just to `signer` - should it ever change we don't have to modify the attribute name

atticusofsparta: [3b23f80](https://github.com/ar-io/ar-io-sdk/pull/175/commits/3b23f80581ea331b89ac31756421f74799315e2c)

dtfiedler (approved):

PR #177, opened 2024-07-30
feat(PE-6516): ant registry class

atticusofsparta:
image

### DIFF
--- a/src/common/contracts/ao-process.ts
+++ b/src/common/contracts/ao-process.ts
@@ -16,10 +16,20 @@
  */
 import { connect } from '@permaweb/aoconnect';
 
-import { AOContract, AoClient, AoSigner } from '../../types.js';
+import {
+  AOContract,
+  AoClient,
+  AoSigner,
+  ProcessConfiguration,
+  isProcessConfiguration,
+  isProcessIdConfiguration,
+} from '../../types.js';
 import { safeDecode } from '../../utils/json.js';
 import { version } from '../../version.js';
-import { WriteInteractionError } from '../error.js';
+import {
+  InvalidContractConfigurationError,
+  WriteInteractionError,
+} from '../error.js';
 import { ILogger, Logger } from '../logger.js';
 
 export class AOProcess implements AOContract {
@@ -39,6 +49,18 @@ export class AOProcess implements AOContract {
     this.processId = processId;
     this.logger = logger;
     this.ao = ao;
+  }
+
+  static fromConfiguration(config: Required<ProcessConfiguration>) {
+    if (isProcessConfiguration(config)) {
+      return config.process;
+    } else if (isProcessIdConfiguration(config)) {
+      return new AOProcess({
+        processId: config.processId,
+      });
+    } else {
+      throw new InvalidContractConfigurationError();
+    }
   }
 
   async read<K>({
--- a/src/common/contracts/ao-process.ts
+++ b/src/common/contracts/ao-process.ts
@@ -16,20 +16,10 @@
  */
 import { connect } from '@permaweb/aoconnect';
 
-import {
-  AOContract,
-  AoClient,
-  AoSigner,
-  ProcessConfiguration,
-  isProcessConfiguration,
-  isProcessIdConfiguration,
-} from '../../types.js';
+import { AOContract, AoClient, AoSigner } from '../../types.js';
 import { safeDecode } from '../../utils/json.js';
 import { version } from '../../version.js';
-import {
-  InvalidContractConfigurationError,
-  WriteInteractionError,
-} from '../error.js';
+import { WriteInteractionError } from '../error.js';
 import { ILogger, Logger } from '../logger.js';
 
 export class AOProcess implements AOContract {
@@ -49,18 +39,6 @@ export class AOProcess implements AOContract {
     this.processId = processId;
     this.logger = logger;
     this.ao = ao;
-  }
-
-  static fromConfiguration(config: Required<ProcessConfiguration>) {
-    if (isProcessConfiguration(config)) {
-      return config.process;
-    } else if (isProcessIdConfiguration(config)) {
-      return new AOProcess({
-        processId: config.processId,
-      });
-    } else {
-      throw new InvalidContractConfigurationError();
-    }
   }
 
   async read<K>({
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -38,7 +38,6 @@ export const ioDevnetProcessId = IO_DEVNET_PROCESS_ID;
 export const IO_TESTNET_PROCESS_ID =
   'agYcCFJtrMG6cqMuZfskIkFTGvUPddICmtQSBIoPdiA';
 
-// TODO: set ant registry ID with deployed contract
 export const ANT_REGISTRY_ID = 'i_le_yKKPVstLTDSmkHRqf-wYphMnwB9OhleiTgMkWc';
 export const MIO_PER_IO = 1_000_000;
 export const AOS_MODULE_ID = 'cbn0KKrBZH7hdNkNokuXLtGryrWM--PjSTBqIzw9Kkk';
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -37,6 +37,10 @@ export const IO_DEVNET_PROCESS_ID =
 export const ioDevnetProcessId = IO_DEVNET_PROCESS_ID;
 export const IO_TESTNET_PROCESS_ID =
   'agYcCFJtrMG6cqMuZfskIkFTGvUPddICmtQSBIoPdiA';
+
+// TODO: set ant registry ID with deployed contract
+export const ANT_REGISTRY_ID =
+  process.env.ANT_REGISTRY_ID ?? 'todo-set-ant-registry-id';
 export const MIO_PER_IO = 1_000_000;
 export const AOS_MODULE_ID = 'cbn0KKrBZH7hdNkNokuXLtGryrWM--PjSTBqIzw9Kkk';
 export const ANT_LUA_ID = 'Flwio4Lr08g6s6uim6lEJNnVGD9ylvz0_aafvpiL8FI';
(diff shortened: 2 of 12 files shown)


atticusofsparta:
Wondering if we want to list this in the README/docs yet. Its 2 functions, register and accessControlList that are on the class. 

Currently not tested, need to deploy an ant registry to do so.

Note the `accessControlList` method comment - should we rename to getANTsByAddress?

src/utils/ao.ts
```diff
@@ -54,6 +57,12 @@ export async function spawnANT({
     module,
     scheduler,
     signer,
+    tags: [
+      {
+        name: 'ANT-Registry-Id',
+        value: antRegistryId,
+      },
+    ],
```
dtfiedler, line 65: what's the intent of this?

src/common/contracts/ao-process.ts
```diff
@@ -41,6 +51,18 @@ export class AOProcess implements AOContract {
...
+    if (isProcessConfiguration(config)) {
+      return config.process;
+    } else if (isProcessIdConfiguration(config)) {
+      return new AOProcess({
+        processId: config.processId,
+      });
+    } else {
+      throw new InvalidContractConfigurationError();
+    }
+  }
```
dtfiedler, line 64: we can refactor this later - revert for now and we can

src/utils/ao.ts
atticusofsparta: Initialization parameters for ant contracts so they can be initialized with a target registry ID

dtfiedler:
please add e2e tests with our existing registry

src/utils/processes.ts
```diff
@@ -31,45 +33,13 @@ import {
...
-  );
-
-  if (ownedOrControlledByWallet.length === 0) {
-    return [];
-  }
-
-  // TODO: insert gql query to find ANT processes owned by wallet given wallet not currently in the registry
-  return [...new Set(ownedOrControlledByWallet)] as string[];
+  const res = await registry.accessControlList({ address });
+  return [...res.Owned, ...res.Controlled];
```
dtfiedler, line 42: nit: use a Set to filter out duplicates

src/utils/processes.ts
```diff
@@ -192,7 +192,7 @@ export const fetchAllArNSRecords = async ({
   do {
     const pageResult = await contract
       .getArNSRecords({ cursor, limit: pageSize })
-      .catch((e: any) => {
+      .catch((e: Error) => {
```
dtfiedler, line 195: prefer any here bc we want to use `e?.message` and `e?.stack` if available without worrying about casting

src/constants.ts
```diff
@@ -37,6 +37,10 @@ export const IO_DEVNET_PROCESS_ID =
 export const ioDevnetProcessId = IO_DEVNET_PROCESS_ID;
 export const IO_TESTNET_PROCESS_ID =
   'agYcCFJtrMG6cqMuZfskIkFTGvUPddICmtQSBIoPdiA';
+
+// TODO: set ant registry ID with deployed contract
+export const ANT_REGISTRY_ID =
```
dtfiedler, line 42: don't support an env var here - let clients decide if they want to use something else

src/utils/json.ts
```diff
@@ -14,6 +14,8 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function safeDecode<R = unknown>(data: any): R {
```
dtfiedler, line 19:
```suggestion
export function safeDecode<R = unknown>(data: unknown): R {
```

atticusofsparta: unknown is not accepted by json.parse

tests/e2e/cjs/index.test.js
```diff
@@ -266,3 +280,29 @@ describe('IO', async () => {
...
+  it('should deploy and register a new ANT', async () => {
+    const antId = await spawnANT({
+      signer,
+    });
+    await registry.register({ processId: antId });
+    await new Promise((resolve) => setTimeout(resolve, 5000));
+    const antIdsRes = await registry.accessControlList({ address });
+    const antIds = [...antIdsRes.Owned, ...antIdsRes.Controlled];
+    assert(antIds[0] == antId);
+  });
```
dtfiedler, line 307: just do a read test here - these tests are to ensure the apis work in the respective environment, we should not actually write any messages to the process from these

src/constants.ts
```diff
@@ -37,6 +37,9 @@ export const IO_DEVNET_PROCESS_ID =
 export const ioDevnetProcessId = IO_DEVNET_PROCESS_ID;
 export const IO_TESTNET_PROCESS_ID =
   'agYcCFJtrMG6cqMuZfskIkFTGvUPddICmtQSBIoPdiA';
+
+// TODO: set ant registry ID with deployed contract
```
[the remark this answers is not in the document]
atticusofsparta: tis not, removed

tests/e2e/cjs/index.test.js
```diff
@@ -266,3 +280,22 @@ describe('IO', async () => {
...
+  let registry;
+  let wallet;
+  let address;
+
+  before(async () => {
+    wallet = await arweave.wallets.generate();
+    address = await arweave.wallets.jwkToAddress(wallet);
+    const arbundlesSigner = new ArweaveSigner(wallet);
+    registry = ANTRegistry.init({ signer: arbundlesSigner });
+  });
```
dtfiedler, line 294: you don't need to do this, just just the registry without the signer

tests/e2e/cjs/index.test.js
```diff
@@ -266,3 +280,22 @@ describe('IO', async () => {
...
+
+  it('should retrieve ids from registry', async () => {
+    const antIdsRes = await registry.accessControlList({ address });
```
dtfiedler, line 297: use a known wallet address (or just any valid arweave address)

tests/e2e/cjs/index.test.js
```diff
@@ -266,3 +266,18 @@ describe('IO', async () => {
...
+describe('ANTRegistry', async () => {
+  let registry;
+  let address = '7waR8v4STuwPnTck1zFVkQqJh5K9q9Zik4Y5-5dV7nk';
+
+  before(async () => {
+    registry = ANTRegistry.init();
+  });
+
+  it('should retrieve ids from registry', async () => {
+    const antIdsRes = await registry.accessControlList({ address });
```
dtfiedler, line 279:
```suggestion
  const registry = ANTRegistry.init();
  const address = '7waR8v4STuwPnTck1zFVkQqJh5K9q9Zik4Y5-5dV7nk';

  it('should retrieve ids from registry', async () => {
    const antIdsRes = await registry.accessControlList({ address });
```

tests/e2e/cjs/index.test.js
```diff
@@ -266,3 +266,18 @@ describe('IO', async () => {
...
+    const antIds = [...antIdsRes.Owned, ...antIdsRes.Controlled];
+    assert(antIds instanceof Array);
+  });
```
dtfiedler, line 282: sigh...think about what we're testing here. we want to validate that the response from the contract is what the SDK expects to return back to clients, so that is what we care about.


```suggestion
  const registry = ANTRegistry.init();
  const address = '7waR8v4STuwPnTck1zFVkQqJh5K9q9Zik4Y5-5dV7nk';
  it('should retrieve ids from registry', async () => {
    const affiliatedAnts = await registry.accessControlList({ address });
    assert(typeof affiliatedAnts.Owned === "array")
    assert(typeof affilatedAnts.Controller === "array")
```

tests/e2e/esm/index.test.js
```diff
@@ -267,3 +267,19 @@ describe('IO', async () => {
...
+  before(async () => {
+    registry = ANTRegistry.init();
+  });
+
+  it('should retrieve ids from registry', async () => {
+    const antIdsRes = await registry.accessControlList({ address });
+    const antIds = [...antIdsRes.Owned, ...antIdsRes.Controlled];
+    assert(antIds instanceof Array);
+  });
+});
```
dtfiedler, line 285: see other comments - please be more intentional here

src/constants.ts
```diff
@@ -37,6 +37,8 @@ export const IO_DEVNET_PROCESS_ID =
 export const ioDevnetProcessId = IO_DEVNET_PROCESS_ID;
 export const IO_TESTNET_PROCESS_ID =
   'agYcCFJtrMG6cqMuZfskIkFTGvUPddICmtQSBIoPdiA';
+
```
dtfiedler, line 40:
```suggestion
```

tests/e2e/web/src/App.test.tsx
```diff
@@ -23,4 +27,11 @@ describe('ESM browser validation', () => {
     // check the sdk loaded the data
     expect(result).toHaveTextContent('true');
   });
+
+  it('should retrieve ids from registry', async () => {
+    const affiliatedAnts = await registry.accessControlList({ address });
+
+    expect(Array.isArray(affiliatedAnts.Owned)).toEqual(true);
+    expect(Array.isArray(affiliatedAnts.Controlled)).toEqual(true);
```
dtfiedler, line 35: the goal of this test should be to validate that the registry component can be used in `App.tsx` - refer to the other test for how to align

dtfiedler (changes requested):
see comments r.e. the web e2e tests

tests/e2e/web/src/App.test.tsx
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/177/commits/38ca913fc9f935663b750e4f68e97879d3b70e5f

dtfiedler: ...this test is not checking the behavior of successful call to the registry on App.tsx - please look at the test above

atticusofsparta: Gotchya, updated to use the render logic.

dtfiedler (approved):

PR #184, opened 2024-08-07
fix(register): update spawn ant to register at end of spawn

atticusofsparta:


src/utils/ao.ts
```diff
@@ -59,6 +59,12 @@ export async function spawnANT({
   stateContractTxId?: string;
   antRegistryId?: string;
 }): Promise<string> {
+  // AoSigner is not a Contract Signer - should probably add that to the contract signer type
```
dtfiedler, line 62: what would we need to support it and use ANTRegistry here?

atticusofsparta: We already generate the ao signer in the contract implementations so it would just set it as the signer if it matches the type instead of creating it internally

dtfiedler (approved):

PR #186, opened 2024-08-19
feat(signing): add ao signer as option

atticusofsparta:


dtfiedler:
I'd expect to see some changes to either `createAoSigner` or where we are using it if `AoSigner` is provided.

example: https://github.com/ar-io/ar-io-sdk/blob/main/src/common/io.ts#L596

additionally, please add `init` e2e tests for that handles an `init` with `ArweaveSigner`, `AoSigner`, etc.

src/utils/ao.ts
```diff
@@ -158,6 +158,10 @@ export async function evolveANT({
 }
 
 export function createAoSigner(signer: ContractSigner): AoSigner {
+  if (typeof signer == 'function') {
+    return signer;
+  }
```
dtfiedler, line 163: nit: why not check if `signer instanceof AoSigner`?

tests/e2e/cjs/index.test.js
```diff
@@ -277,3 +287,53 @@ describe('ANTRegistry', async () => {
...
+      const io = IO.init({ signer });
+      const res = await io
+        .transfer({
+          target: 'QGWqtJdLLgm2ehFWiiPzMaoFLD50CnGuzZIPEdoDRGQ',
+          qty: 1,
+        })
+        .catch((e) => e);
+      // if it is a WriteInteractionError, it means the transaction was signed contract threw an error
+      assert(res instanceof WriteInteractionError);
+    }
```
dtfiedler, line 309: we should not be sending any real messages - the tests should only be validating the `init` function that accepts the signer and the expected instance type is returned.

dtfiedler (changes requested):
remove actual writing of messages and check nit comments related to `createAoSigner`

src/utils/ao.ts
atticusofsparta: I thought the same, but AoSigner is a type, not a class

tests/utils.ts
```diff
@@ -2,138 +2,9 @@ import Arweave from 'arweave';
...
-    }),
-    evaluationManifest: {
-      evaluationOptions: {
-        sourceType: SourceType.ARWEAVE,
-      },
-    },
-  });
-}
-
-export async function deployArIOContract({
```
dtfiedler, line 50: 👍 - good call, please remove any smartweave cruft you find

tests/e2e/cjs/index.test.js
```diff
@@ -277,3 +286,38 @@ describe('ANTRegistry', async () => {
...
+  before(async () => {
+    const jwk = await arweave.wallets.generate();
+    signers = [new ArweaveSigner(jwk), createAoSigner(new ArweaveSigner(jwk))];
+  });
+
+  it('Should be able to sign on the IO contract with all ContractSigner types', async () => {
+    for (const signer of signers) {
+      const io = IO.init({ signer });
+
+      assert(io);
```
dtfiedler, line 301: instance of IOWriteable

src/utils/ao.ts
```diff
@@ -157,7 +158,43 @@ export async function evolveANT({
   return id;
 }
 
+export function isAoSigner(value: unknown): value is AoSigner {
+  const TagSchema = z.object({
+    name: z.string(),
+    value: z.union([z.string(), z.number()]),
+  });
```
dtfiedler, line 165: does aoconnect expose any of these schemas?

tests/e2e/esm/index.test.js
```diff
@@ -278,3 +287,38 @@ describe('ANTRegistry', async () => {
     assert(Array.isArray(affiliatedAnts.Controlled));
   });
 });
+
+describe('Signing', async () => {
+  let signers = [];
+  before(async () => {
+    const jwk = await arweave.wallets.generate();
```
dtfiedler, line 294: nit: this is an expensive operation and creates side effects - can we just load a test wallet into the repo and use the same one every time?

src/utils/ao.ts
atticusofsparta: It has them but doesn't expose them - and even if it did, they are inaccurate. They claim tag values can only be strings, and they use `any` for the data type.

dtfiedler: FWIW tags must be converted to strings when posting to arweave, so I wouldn't go sofar as to say their types are inaccurate

tests/e2e/cjs/index.test.js
```diff
@@ -277,3 +290,37 @@ describe('ANTRegistry', async () => {
     assert(Array.isArray(affiliatedAnts.Controlled));
   });
 });
+
+describe('Signing', async () => {
```
dtfiedler, line 294: related to test structure this `describe` is inaccurate. we're really just confirming the signers provided to IO are supported, so I wouldn't place the respective `its` under existing `desribes` for each top level class we export

tests/e2e/cjs/index.test.js
```diff
@@ -277,3 +290,37 @@ describe('ANTRegistry', async () => {
...
+    createAoSigner(new ArweaveSigner(testWallet)),
+  ];
+
+  it('Should be able to sign on the IO contract with all ContractSigner types', async () => {
+    for (const signer of signers) {
+      const io = IO.init({ signer });
+
+      assert(io instanceof IOWriteable);
+    }
+  });
```
dtfiedler, line 306: move under `describe("IO"` and tests are typically lower case (e.g. `it('should be able to sign..',`

dtfiedler (approved):

PR #189, opened 2024-08-26
feat(PE-6579): add AoAntState typeguard util

atticusofsparta:


src/utils/ao.ts
```diff
@@ -218,3 +219,34 @@ export function createAoSigner(signer: ContractSigner): AoSigner {
...
+            transactionId: z.string(),
+            ttlSeconds: z.number(),
+          })
+          .passthrough(),
+      ),
+      Balances: z.record(z.string(), z.number()),
+    })
+    .passthrough();
+
+  AntStateSchema.parse(state);
```
dtfiedler, line 249: nit: may be better to catch and throw false here?

src/utils/ao.ts
```diff
@@ -218,3 +219,40 @@ export function createAoSigner(signer: ContractSigner): AoSigner {
...
+ * @throws {z.ZodError} if the state object does not match the expected schema
+ */
+export function isAoANTState(state: object): state is AoANTState {
+  try {
+    AntStateSchema.parse(state);
+
+    return true;
+  } catch (error) {
+    // this allows us to see the path of the error in the object as well as the expected schema on invalid fields
+    Logger.default.error(error.issues);
```
dtfiedler, line 255: nit: pass a logger as an function parameter (you can default it to Logger.default)

dtfiedler (approved):

PR #220, opened 2024-10-04
fix(PE-6890): enable logging in spawn and evolve utils

atticusofsparta:


dtfiedler (approved):

src/utils/ao.ts
```diff
@@ -43,6 +44,8 @@ export async function spawnANT({
   state,
   stateContractTxId,
   antRegistryId = ANT_REGISTRY_ID,
+  logger = Logger.default,
```
dtfiedler, line 47: doesn't look like you're importing Logger?

PR #221, opened 2024-10-04
feat(PE-6824): add gateway class for interacting with ario gateways

atticusofsparta:


dtfiedler:
moving to draft - need to discuss more what the usefulness of this is

dtfiedler:
Closing for now - this was more intended to be a class for `Gateways` in the gar, and moving things like `updateSettings` APIs there - with the IO contract as a dependency. We can come back to it another time.

PR #227, opened 2024-10-11
fix(PE-6929): validate ant read

atticusofsparta:


tests/unit/token.test.ts
```diff
@@ -1,7 +1,7 @@
 import { strict as assert } from 'node:assert';
 import { describe, it } from 'node:test';
 
-import { IOToken, mIOToken } from '../../src/token.js';
+import { IOToken, mIOToken } from '../../src/types/token';
```
dtfiedler, line 4: nit: continue to use `.js` import paths

src/types/ant.ts
```diff
@@ -0,0 +1,165 @@
...
+    options?: WriteOptions,
+  ): Promise<AoMessageResult>;
+}
+
+export const AntRecordSchema = z
+  .object({
+    transactionId: z.string(),
+    ttlSeconds: z.number(),
+  })
+  .passthrough();
```
dtfiedler, line 117: zod allows inferring types from schemas - so we should replace our existing types with ones inferred on the zod schemas for ANT specific types in this PR. 

i.e. replace `AoANTRecord` type above with 

```typescript
export type AoANTRecord = z.infer(typeof AntRecordSchema)
```

atticusofsparta: ME LIKEY

src/common/ant.ts
```diff
@@ -76,6 +86,13 @@ export class AoANTReadable implements AoANTRead {
     const res = await this.process.read<AoANTState>({
       tags,
     });
+    const schemaResult = AntStateSchema.safeParse(res);
```
atticusofsparta, line 89: Looking at this I wonder if we want to be able provide a `schema` and `errorSchema` to process.read and process.write to enable tapping into the configured logger as well as configuring behaviour on bad results.

We have some specific patterns on error in the ANT and ArNS processes that might be good to enforce.

src/types/ant.ts
```diff
@@ -0,0 +1,177 @@
...
+        "code": "custom",
+        "message": "Must be an Arweave Transaction ID",
+        "path": [
+          "Records",
+          "record1",
+          "transactionId"
+        ]
+      },
+ */
+export const ArweaveTxIdSchema = z
```
atticusofsparta, line 34: We dont current support eth or solana address in our ants, if/when we do this can be made into a union

src/types/common.ts
```diff
@@ -88,3 +88,22 @@ export interface AoClient {
   unmonitor: typeof unmonitor;
   dryrun: typeof dryrun;
 }
+
+export interface AOContract {
```
atticusofsparta, line 92: in the constructor we could pass in a response parser - for example, an ANTResponseParser class or function that has a map for validating happy and sad paths.

Would prevent having to specify parsing of schemas in each method on implementations

atticusofsparta: I refactored - but did not use the data attribute. In this particular case I think the JSX approach is cleaner.

If I had a non binary value that would require a multiline JSX case then I would probably use the data attribute though. Pretty cool that tailwind allows that.

src/common/ant.ts
```diff
@@ -76,6 +86,19 @@ export class AoANTReadable implements AoANTRead {
...
+          Records: z.record(z.string(), AntRecordSchema.passthrough()),
+        }),
+      )
+      .safeParse(res);
+    if (!schemaResult.success) {
+      throw new Error(
+        'Invalid ANT State\n' +
+          JSON.stringify(schemaResult.error.format(), null, 2),
+      );
+    }
```
dtfiedler, line 101: nit, this could be a small utility function to format the error - either as a utility or private function on the class

src/types/ant.ts
```diff
@@ -0,0 +1,195 @@
...
+  logger: Logger = Logger.default,
+): state is AoANTState {
+  try {
+    AntStateSchema.parse(state);
+    return true;
+  } catch (error) {
+    // this allows us to see the path of the error in the object as well as the expected schema on invalid fields
+    logger.error(error.issues);
+    return false;
+  }
```
dtfiedler, line 134: nit - use `safeParse` and return `result.success`

dtfiedler (changes requested):
a few nits, and a rebase with alpha required

src/common/ant.ts
dtfiedler: i'd prefer this over providing it to the constructor, but it pushes down the error handling to a class that we should avoid over prescribing. I'd rather these exist on the higher level classes `ANT` and `IO` so we can responsibly handle separately from AO specific errors

src/types/common.ts
dtfiedler: see comment above

src/types/ant.ts
```diff
@@ -0,0 +1,186 @@
...
+  ),
+  Ticker: z.string().describe('The ticker symbol for the ANT.'),
+  ['Total-Supply']: IntegerStringSchema.describe(
+    'Total supply of the ANT in circulation.',
+  ),
+  Logo: ArweaveTxIdSchema.describe('Transaction ID of the ANT logo.'),
+  Denomination: IntegerStringSchema.describe(
+    'The number of decimal places to use for the ANT. Defaults to 0 if not set representing whole numbers.',
+  ),
+  HandlerNames: z
```
dtfiedler, line 113: let's modify this in the ants to be `Handlers` to match the ar-io-network-process

atticusofsparta: _I found 2 ants with different names for the handlers. It is in fact currently "Handlers". We updated it couple weeks ago and now we have old ants with "HandlerNames" and new ones with "Handlers". I currently set them both to optional and marked the HandlerNames one as deprecated in the description for the schema field. Zod doesnt have a built in way to show deprecations there and dont think we want to blow up here.
This is making me wonder if maybe we should have a flag in the constructor for "strict" apis that do blow up on bad data. Or maybe something like ant.strict.getState()_

atticusofsparta: I found 2 ants with different names for the handlers. It is in fact currently "Handlers". We updated it couple weeks ago and now we have old ants with "HandlerNames" and new ones with "Handlers". I currently set them both to optional and marked the HandlerNames one as deprecated in the description for the schema field. Zod doesnt have a built in way to show deprecations there and dont think we want to blow up here.
This is making me wonder if maybe we should have a flag in the constructor for "strict" apis that do blow up on bad data. Or maybe something like ant.strict.getState()

src/utils/index.ts
dtfiedler, line 1: we should generally start avoiding the use of these barrel files - they make tree shaking difficult and ultimately mask correct import paths. not a necessary change for now, but just to keep in mind

tests/e2e/cjs/index.test.js
```diff
@@ -359,14 +361,73 @@ describe('ANTRegistry', async () => {
...
+    assert(balance);
+    const balances = await ant.getBalances({ strict });
+    assert(balances);
+    const controllers = await ant.getControllers({ strict });
+    assert(controllers);
+    const records = await ant.getRecords({ strict });
+    assert(records);
+    const record = await ant.getRecord({ undername: 'ardrive' }, { strict });
+    assert(record);
+  });
```
dtfiedler, line 432: nit, move these to just the esm tests - we have the coverage we need for CJS (really just validating the client imports and instantion works) - we can focus the full suite of integration tests on the `esm` tests without duplicating

tests/e2e/esm/index.test.js
```diff
@@ -396,34 +396,41 @@ describe('ANTRegistry', async () => {
...
 
   it('should be able to get ANT info', async () => {
-    const info = await ant.getInfo({ processId });
+    const info = await ant.getInfo({ strict: true });
     assert.ok(info);
   });
 
   it('should be able to get the ANT records', async () => {
-    const records = await ant.getRecords({ processId });
+    const records = await ant.getRecords({ strict: true }, { strict: true });
```
dtfiedler, line 428:
```suggestion
    const records = await ant.getRecords({ strict: true });
```

tests/e2e/esm/index.test.js
```diff
@@ -396,34 +396,41 @@ describe('ANTRegistry', async () => {
...
 
-      assert(writeable instanceof AoANTWriteable);
+      assert(nonStrictAnt instanceof AoANTWriteable);
+      assert(strictAnt instanceof AoANTWriteable);
     }
   });
 
   it('should be able to get ANT info', async () => {
-    const info = await ant.getInfo({ processId });
+    const info = await ant.getInfo({ strict: true });
```
dtfiedler, line 423: Note: we don't need all these tests to enforce strict on the APIs. The main thing to test would be overriding the `strict` defined on the constructor to ensure the flexibility works as expected.

atticusofsparta: [62c9140](https://github.com/ar-io/ar-io-sdk/pull/227/commits/62c914086b936b0c91cea757709941c1987f3ade)

tests/e2e/esm/index.test.js
atticusofsparta: [62c9140](https://github.com/ar-io/ar-io-sdk/pull/227/commits/62c914086b936b0c91cea757709941c1987f3ade)

dtfiedler (approved):

PR #230, opened 2024-10-15
feat(arns): add `ArNSResolver` class that enables forward and reverse name resolution

dtfiedler:
This includes a node-cache dep, we can just use an in-memory map and introduce other deps via plugins to avoid bolting this in the SDK even if clients do not need it


dtfiedler:
Closing for now

PR #248, opened 2024-11-07
fix(PE-7080): add apis for fetching paginated delegates

dtfiedler:
TODO: 
- apis for updating delegates (ref: https://github.com/ar-io/ar-io-network-process/blob/develop/src/main.lua#L2913-L3009)

### DIFF
--- a/src/common/io.ts
+++ b/src/common/io.ts
@@ -213,14 +213,14 @@ export class IOReadable implements AoIORead {
   }
 
   async getArNSRecords(
-    pageParams?: PaginationParams,
+    params?: PaginationParams<AoArNSNameDataWithName>,
   ): Promise<PaginationResult<AoArNSNameDataWithName>> {
     const allTags = [
       { name: 'Action', value: 'Paginated-Records' },
-      { name: 'Cursor', value: pageParams?.cursor?.toString() },
-      { name: 'Limit', value: pageParams?.limit?.toString() },
-      { name: 'Sort-By', value: pageParams?.sortBy },
-      { name: 'Sort-Order', value: pageParams?.sortOrder },
+      { name: 'Cursor', value: params?.cursor?.toString() },
+      { name: 'Limit', value: params?.limit?.toString() },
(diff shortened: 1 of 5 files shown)


src/common/io.ts
```diff
@@ -313,6 +287,40 @@ export class IOReadable implements AoIORead {
     });
   }
 
+  async getGatewayDelegates(
+    params?: { address: WalletAddress } & PaginationParams,
```
arielmelendez, line 291: Would be interesting if the paginationparams type were genericized to allow for specifying the allowable sortBy values for each use case.

fedellen (dismissed):

src/types/io.ts
```diff
@@ -312,6 +315,14 @@ export interface AoIORead {
   }: {
     address: WalletAddress;
   }): Promise<AoGateway | undefined>;
+  // TODO: these could be moved to a separate Gateways class that implements gateway specific interactions
+  getGatewayDelegates(
+    params?: PaginationParams,
```
arielmelendez, line 320: The input type here is different the actual implementation (where `address` is unioned into the type). Same below.

tests/unit/utils.test.ts
```diff
@@ -0,0 +1,44 @@
...
+      { name: 'Tag1', value: undefined },
+      { name: 'Tag2', value: undefined },
+    ];
+
+    const prunedTags = pruneTags(tags);
+
+    assert.deepEqual(prunedTags, []);
+  });
+
+  it('should return same array when no tags have undefined values', () => {
```
arielmelendez, line 34: Could add an empty tags test as well.

dtfiedler: [7a1abc4](https://github.com/ar-io/ar-io-sdk/pull/248/commits/7a1abc4869aed61a39cd99a313c64aff7c5e10de)

fedellen (approved):

arielmelendez (approved):

PR #249, opened 2024-11-07
chore(util): add utility for fetching block timestamp

dtfiedler:


tests/unit/utils.test.ts
```diff
@@ -48,3 +52,13 @@ describe('pruneTags', () => {
...
   });
 });
+
+describe('getCurrentBlockUnixTimestamp', () => {
+  it('should return the current block timestamp', async () => {
+    // stub arweave block request using nock
+    const arweave = Arweave.init({});
+    const minTimestamp = Date.now();
+    const timestamp = await getCurrentBlockUnixTimestamp(arweave);
+    assert.ok(timestamp >= minTimestamp);
```
fedellen, line 62: where does `nock` effect this?

src/utils/arweave.ts
```diff
@@ -34,3 +36,16 @@ export const pruneTags = (
...
+  arweave: Arweave,
+): Promise<Timestamp> => {
+  return await arweave.blocks
+    .getCurrent()
+    .then((block) => {
+      return block.timestamp * 1000;
+    })
+    .catch(() => {
+      return Date.now(); // fallback to current time
+    });
```
fedellen, line 50: nit: I get wary of `return await` by itself, but adding a .then+.catch handler on top of it makes it harder to reason -- and it swallows potential errors the client might want to know. could just be like 
```suggestion
  try {
    const block = await arweave.blocks
    .getCurrent();
    return block.timestamp * 1000
  } catch { // fallback to current time -- consider logging error and fallback intent
    return Date.now();
  }
```

tests/unit/utils.test.ts
dtfiedler: i'm not using nock atm

src/utils/arweave.ts
dtfiedler: these are inconsequential errors - a logger could be passed in to the function.

There is a global `Logger.default` available for use, not too dissimilar from turbos setup.

tests/unit/utils.test.ts
dtfiedler: will remove the comment

src/utils/arweave.ts
dtfiedler: also r.e. try/catch. i understand your point. this is more concise but less readable, fair. but we're not doing anything nuanced here. imo the try/catch pattern is actually HARDER to follow in most scenarios (especially with massive try blocks)

src/utils/arweave.ts
```diff
@@ -34,3 +36,16 @@ export const pruneTags = (
     }): tag is { name: string; value: string } => tag.value !== undefined,
   );
 };
+
+export const getCurrentBlockUnixTimestamp = async (
```
atticusofsparta, line 40: nit: this returns a js timestamp, not unix

dtfiedler: [fa29090](https://github.com/ar-io/ar-io-sdk/pull/249/commits/fa290906d9ca5c02fd43cf7eda43f708e8792694)

tests/unit/utils.test.ts
dtfiedler: [c98de0f](https://github.com/ar-io/ar-io-sdk/pull/249/commits/c98de0f11e81631cea5ebe1cac310dc61074353e)

PR #253, opened 2024-11-08
fix(exports): update export files

atticusofsparta:


src/common/index.ts
```diff
@@ -17,6 +17,7 @@ export * from './error.js';
 export * from './logger.js';
 export * from './ant.js';
 export * from './ant-registry.js';
+export * from './http.js';
```
dtfiedler, line 20: do we need to export this?

atticusofsparta: Actually no we don't - its an artifact from arns-service days. Worth leaving as we can use it for the future gateway class (matter of fact i think i have it implemented in a draft pr) but no need to export it for the moment. 

https://github.com/ar-io/ar-io-sdk/pull/253/commits/d6369aa0a722abbe468eb0e73423b1b757dbf0ee

fedellen (approved):

PR #259, opened 2024-11-13
feat(PE-7107): update getRecords to return as flat array of objects

atticusofsparta:


dtfiedler (changes requested):

tests/e2e/esm/index.test.js
```diff
@@ -544,7 +544,8 @@ describe('e2e esm tests', async () => {
   });
 
   describe('ANT', async () => {
-    const processId = 'YcxE5IbqZYK72H64ELoysxiJ-0wb36deYPv55wgl8xo';
+    // ANT v8 source
+    const processId = 'oQ4GNTed8cnNw-H5olq606gCFd5MbGSV4NTmfpIW4FI';
```
dtfiedler, line 548: nit: let's add a test for an older ant as well - can be in just a single `it` block

atticusofsparta: [0489cb6](https://github.com/ar-io/ar-io-sdk/pull/259/commits/0489cb6924b68c19b8791cc47e0f7c1589be842f) - just a single getInfo call

dtfiedler: it should be validating the getRecords api

tests/e2e/esm/index.test.js
```diff
@@ -544,14 +544,27 @@ describe('e2e esm tests', async () => {
...
     const ant = ANT.init({
       process: new AOProcess({
         processId,
         ao: aoClient,
       }),
     });
 
+    it('should be able to get info on old ant', async () => {
```
dtfiedler, line 556: we changed the getRecords API - we want to test that the SDK correctly maps an old ANTs record response...nothing has changed about the info handler - so this isn't validating anything

atticusofsparta: [77601b2](https://github.com/ar-io/ar-io-sdk/pull/259/commits/77601b2fc9117a2a4b9ff1226fa3f329e8a74e1d)

dtfiedler (approved):

PR #265, opened 2024-11-18
fix(arns): update reserved names to pagaination api

dtfiedler:
Technically breaking - but should only impact arns.app


atticusofsparta (approved):
req e2e test for reserved names

PR #273, opened 2024-11-21
feat(redelegate stake): init IO methods PE-7159

fedellen:
this PR adds the Redelegate Stake write action and get redelegation fee read methods to the IO class

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,8 @@ This is the home of [ar.io] SDK. This SDK provides functionality for interacting
     - [`getPrimaryNames({ cursor, limit, sortBy, sortOrder })`](#getprimarynames-cursor-limit-sortby-sortorder-)
     - [`getPrimaryName({ name, address })`](#getprimaryname-name-address-)
     - [`requestPrimaryName({ name, address })`](#requestprimaryname-name-address-)
+    - [`redelegateStake({ target, source, stakeQty, vaultId })`](#redelegatestake-target-source-stakeqty-vaultid-)
+    - [`getRedelegationFee({ address })`](#getredelegationfee-address-)
   - [Configuration](#configuration)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1591,6 +1593,66 @@ const { id: txId } = await io.requestPrimaryName({
...
+
+<details>
+  <summary>Output</summary>
+
+```json
+{
+  "sourceGateway": {
+    // ...source gateway details
+  },
+  "targetGateway": {
```
dtfiedler, line 1621: do we return the full gateway here? should we return just the addresses?

dtfiedler: more concerned about the Handler returning `vaults` and `delegates` with large state. let's make sure those at least are removed from the handler call

dtfiedler: another note - for write messages, we return the ID of the response, not the entire result object. so this actually is not the correct output for this API, can remove.

the above comment still applies about what the handler returns. i think just the addresses are suffficent.

README.md
```diff
@@ -1591,6 +1593,66 @@ const { id: txId } = await io.requestPrimaryName({
...
+  "feeResetTimestamp": 1730996691117,
+  "redelegationsSinceFeeReset": 1
+}
+```
+
+</details>
+
+#### `getRedelegationFee({ address })`
+
+Retrieves the fee rate as percentage required to redelegate the stake of a specific address.
```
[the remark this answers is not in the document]
fedellen: its 0% for first, 10% for second, 20% for third, and so on until it caps at 60%: f438bb8

README.md
fedellen: removed the output here, to match the other write actions:
[28edd4e](https://github.com/ar-io/ar-io-sdk/pull/273/commits/28edd4e74ca76c270f3bce4a62434aa49335dbdb)

dtfiedler (approved):

PR #275, opened 2024-11-22
fix(ant): revert breaking change on records for ANT

dtfiedler:


atticusofsparta:
Readme should need an update as well.

atticusofsparta (approved):

PR #293, opened 2024-12-02
fix: use `Keywords` for setKeywords

fedellen:


arielmelendez (approved):

atticusofsparta (approved):
Rawer, we need better tooling so we can do integration tests for these.

PR #311, opened 2024-12-10
fix(eth signer): use a unique anchor in ans-104 headers

fedellen:
this will produce non-deterministic IDs for all signer types that dont provide salt

PE-7158

### DIFF
--- a/src/common/contracts/ao-process.ts
+++ b/src/common/contracts/ao-process.ts
@@ -130,6 +130,11 @@ export class AOProcess implements AOContract {
         });
 
         // TODO: do a read as a dry run to check if the process supports the action
+        const randomLetter = String.fromCharCode(
+          65 + Math.floor(Math.random() * 26),
+        );
+        // anchor is a random string produce non-deterministic messages IDs
+        const anchor = Date.now().toString().padEnd(32, randomLetter);
 
         const messageId = await this.ao.message({
           process: this.processId,
@@ -137,11 +142,13 @@ export class AOProcess implements AOContract {
           tags: [...tags, { name: 'AR-IO-SDK', value: version }],
           data,
           signer,
+          anchor,
         });
 
         this.logger.debug(`Sent message to process`, {
           messageId,
           processId: this.processId,
+          anchor,
         });
 
         // check the result of the send interaction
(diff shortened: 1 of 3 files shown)


src/common/contracts/ao-process.ts
```diff
@@ -130,18 +130,25 @@ export class AOProcess implements AOContract {
         });
 
         // TODO: do a read as a dry run to check if the process supports the action
+        const randomLetter = String.fromCharCode(
+          65 + Math.floor(Math.random() * 26),
+        );
+        // anchor is a random string produce non-deterministic messages IDs
+        const anchor = Date.now().toString().padEnd(32, randomLetter);
```
arielmelendez, line 137: This would only let users send basic messages up to 26 times, right? Maybe just use fully random bytes or some combination of the current timestamp plus random bytes. Here's a way to produce random characters:
```
function getRandomText(length = 32) {
  const array = new Uint8Array(length);
  crypto.getRandomValues(array);
  return Array.from(array, byte => byte.toString(16).padStart(2, '0')).join('').slice(0, length);
}
```
crypto.getRandomValues generates cryptographically secure random values.
Array.from with .toString(16): Converts each byte to a hexadecimal string (base 16).
.padStart(2, '0'): Ensures each byte is represented as two characters (e.g., 0a instead of a).
.slice(0, length): Ensures the resulting string is exactly the specified length.

fedellen: was trying to avoid putting `crypto` import here as we don't access it much in the SDK. maybe something like this?

```ts
export function randomString(length: number): string {
  const chars =
    'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
  let result = '';
  for (let i = 0; i < length; i++) {
    result += chars.charAt(Math.floor(Math.random() * chars.length));
  }
  return result;
}
```

arielmelendez: That won't perform as well, I believe, and is less likely to distribute randomness uniformly. If we're already relying on crypto elsewhere and if it won't create additional integration stresses for SDK clients, I recommend we use it.

arielmelendez (approved):

PR #313, opened 2024-12-11
feat(cost-details): init get-cost-details PE-7114

fedellen:


README.md
```diff
@@ -1200,6 +1201,43 @@ const price = await ario
 
 </details>
 
+#### `getCostDetails({ intent, fromAddress, fundFrom, ...args})`
+
+Calculates the expanded cost details for the interaction in question, e.g a 'Buy-record' interaction, where args are the specific params for that interaction. The fromAddress is the address that would be charged for the interaction, and fundFrom is where the funds would be taken from, either `balance`, `stakes`, or `any`.
```
atticusofsparta, line 1206:
```suggestion
Calculates the expanded cost details for the interaction in question, e.g a 'Buy-Record' interaction, where args are the specific params for that interaction. The fromAddress is the address that would be charged for the interaction, and fundFrom is where the funds would be taken from, either `balance`, `stakes`, or `any`.
```

README.md
```diff
@@ -1200,6 +1201,43 @@ const price = await ario
...
+
+```typescript
+const costDetails = await ario.getCostDetails({
+  intent: 'Buy-Record',
+  fromAddress: 't4Xr0_J4Iurt7caNST02cMotaz2FIbWQ4Kbj616RHl3',
+  fundFrom: 'stakes',
+  name: 'ar.io',
```
atticusofsparta, line 1213:
```suggestion
  name: 'ar-io',
```

src/common/contracts/ao-process.ts
```diff
@@ -44,9 +44,11 @@ export class AOProcess implements AOContract {
   async read<K>({
     tags,
     retries = 3,
+    Owner,
   }: {
     tags?: Array<{ name: string; value: string }>;
     retries?: number;
+    Owner?: string;
```
atticusofsparta, line 51: wondering if we should do something like Arguments<this.ao.dryrun>[0] and pass those all thru.

In any case, I suspect what you want here may be `From`, not `Owner` - right?

src/types/io.ts
```diff
@@ -349,6 +349,43 @@ export type AoTokenCostParams = {
   years?: number;
   name: string;
   quantity?: number;
+  fromAddress?: WalletAddress;
+};
+
+export const fundFromOptions = ['balance', 'stakes', 'any'];
+export type FundFrom = (typeof fundFromOptions)[number];
```
atticusofsparta, line 356: Does this give exact types and not just `string` without `as const` on fundFromOptions?

fedellen: seems the auto-fill works when setup like his, but the IDE hover shows the type `string`

fedellen: 3e76ec4

src/common/contracts/ao-process.ts
fedellen: yeah but the type on dryrun only has `Owner` and it seems to do what we need. previously we had no way of indicating the From address on a read action
image

fedellen: exposed as fromAddress in our API fwiw: ee4a024

atticusofsparta: Interesting - it must be adding it on its own if no From param is provided. 

This may have issues if attempting to do a dry run as an ANT (because of authorities) - but this is probably fine for now.

atticusofsparta (approved):

PR #316, opened 2024-12-14
chore(test): update e2e test to leverage typescript

dtfiedler:
This should help sniff out type issues before releases.


tests/e2e/esm/index.test.ts
```diff
@@ -382,7 +366,7 @@ describe('e2e esm tests', async () => {
 
     it('should be able to get balances of a specific to first page', async () => {
       const balances = await ario.getBalances({
-        cursor: 1000000,
+        cursor: '1000000',
```
atticusofsparta, line 369: im a little confused by this test - this cursor seems configured to balances nots addresses?

atticusofsparta (approved):

tests/e2e/esm/index.test.ts
dtfiedler: Balances are a valid cursor, but the type type expects a string.

PR #325, opened 2024-12-19
feat: remove usage of Timestamp in epoch inputs PE-7338

fedellen:


src/common/io.ts
```diff
@@ -154,38 +154,52 @@ export class ARIOReadable implements AoARIORead {
...
+  private async computeEpochIndexForTimestamp(
+    timestamp: number,
+  ): Promise<number> {
+    const epochSettings = await this.getEpochSettings();
+    const epochZeroStartTimestamp = epochSettings.epochZeroStartTimestamp;
+    const epochLengthMs = epochSettings.durationMs;
+    return Math.floor((timestamp - epochZeroStartTimestamp) / epochLengthMs);
+  }
+
+  private async epochInputToTags(
```
fedellen, line 166: nit: can return just one tag for no concat

fedellen: gets complex cause we need to type check undefined, empty array is simpler. think its fine as is

src/common/io.ts
```diff
@@ -154,38 +154,48 @@ export class ARIOReadable implements AoARIORead {
     });
   }
 
+  private async computeEpochIndexForTimestamp(
+    timestamp: number,
+  ): Promise<number> {
+    const epochSettings = await this.getEpochSettings();
```
dtfiedler, line 160: you can memoize this by adding `private epochSettings: AoEpochSettings | undefined`. set it if it has not been set, return it if it already has.

src/common/io.ts
```diff
@@ -409,16 +392,7 @@ export class ARIOReadable implements AoARIORead {
...
-        name: 'Timestamp',
-        value:
-          (epoch as { timestamp?: number })?.timestamp?.toString() ??
-          Date.now().toString(),
-      },
-      {
-        name: 'Epoch-Index',
-        value: (epoch as { epochIndex?: number })?.epochIndex?.toString(),
-      },
+      await this.epochInputToEpochIndexTag(epoch),
```
dtfiedler, line 395: nit: pull this out of all tags to a var

src/common/io.ts
```diff
@@ -154,38 +154,48 @@ export class ARIOReadable implements AoARIORead {
...
+    }
+    const timestamp = (params as { timestamp?: number })?.timestamp;
+    if (timestamp !== undefined) {
+      return {
+        name: 'Epoch-Index',
+        value: (await this.computeEpochIndexForTimestamp(timestamp)).toString(),
+      };
+    }
+    return { name: 'Epoch-Index', value: '' };
+  }
```
dtfiedler, line 182:
```suggestion
  private async computeEpochIndex(
    params?: EpochInput,
  ): Promise<number | undefined> {
   const { epochIndex = undefined, timestamp = undefined } = params; // you can just ts-ignore
    if (timestamp){
        epochIndex ?= await await this.computeEpochIndexForTimestamp(timestamp)).toString()
    }
    return epochIndex
```
and then keep the `Epoch-Index` tag setup in the caller

src/common/io.ts
dtfiedler: correct, in that case it would be a forced refresh

src/common/io.ts
```diff
@@ -154,39 +154,54 @@ export class ARIOReadable implements AoARIORead {
...
-        name: 'Timestamp',
-        value:
-          (params as { timestamp?: number })?.timestamp?.toString() ??
-          Date.now().toString(),
-      },
-      {
-        name: 'Epoch-Index',
-        value: (params as { epochIndex?: number })?.epochIndex?.toString(),
-      },
+      epochInputTag,
```
dtfiedler, line 192: i don't believe this handler supports epoch-index, we don't store settings state on the epoch

fedellen: [b311ca](https://github.com/ar-io/ar-io-sdk/pull/325/commits/1b311ca4d21c33ca5a6a44b942fb5e5673c09c37)

src/common/io.ts
fedellen: [42399a4](https://github.com/ar-io/ar-io-sdk/pull/325/commits/42399a42acc96b796edff30d03824034ea1c3e58)

src/common/io.ts
fedellen: memoized above getEpochSettings

src/common/io.ts
```diff
@@ -154,38 +154,50 @@ export class ARIOReadable implements AoARIORead {
...
+
+    const timestamp = (params as { timestamp?: number })?.timestamp;
+    if (timestamp !== undefined) {
+      return (await this.computeEpochIndexForTimestamp(timestamp)).toString();
+    }
+
+    return undefined;
+  }
+
+  private epochSettings: AoEpochSettings | undefined;
```
dtfiedler, line 182: nit, put this at the top of the class with other attributes - it gets lost down here

src/common/io.ts
```diff
@@ -154,38 +154,50 @@ export class ARIOReadable implements AoARIORead {
...
+    }
+
+    const allTags = [{ name: 'Action', value: 'Epoch-Settings' }];
+
+    const epochSettings = await this.process.read<AoEpochSettings>({
       tags: pruneTags(allTags),
     });
+    this.epochSettings = epochSettings;
+    return epochSettings;
   }
```
dtfiedler, line 195:
```suggestion
    this.epochSettings = await this.process.read<AoEpochSettings>({
      tags: pruneTags(allTags),
    });
    return this.epochSettings;
  }
```

dtfiedler:
do we have e2e tests for these?

dtfiedler (approved):

dtfiedler (approved):

PR #330, opened 2025-01-03
feat(ario): updated tags, add notifyOwners to ANT.removePrimaryNames

dtfiedler:
related - https://github.com/ar-io/ar-io-network-process/pull/296#pullrequestreview-2529230832

### DIFF
--- a/src/types/io.ts
+++ b/src/types/io.ts
@@ -636,21 +636,11 @@ export interface AoARIOWrite extends AoARIORead {
     options?: WriteOptions,
   ): Promise<AoMessageResult>;
   // END OF GATEWAY SPECIFIC INTERACTIONS
-  /* @deprecated - use `buyArNSName` instead */
   buyRecord(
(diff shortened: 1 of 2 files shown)


src/types/io.ts
```diff
@@ -628,14 +628,24 @@ export interface AoARIOWrite extends AoARIORead {
     options?: WriteOptions,
   ): Promise<AoMessageResult>;
   // END OF GATEWAY SPECIFIC INTERACTIONS
+  // @deprecated - use `buyArNSName` instead
```
fedellen, line 631: tsdoc syntax will provide IDE hinting and ~~strikethrough~~

```suggestion
  /** @deprecated - use `buyArNSName` instead */
```

dtfiedler: [30c2e51](https://github.com/ar-io/ar-io-sdk/pull/330/commits/30c2e5111fe8d34d96ca9e002074ca24f37edd57)

PR #337, opened 2025-01-13
feat(io): fetch historical epoch data from gql

dtfiedler:
The process now posts `Epoch-Distribution-Notice` messages with full distributed epoch data. We can fetch these via gql/gateway instead of persisting them in process state.

Related - https://github.com/ar-io/ar-io-network-process/pull/322

### DIFF
--- a/src/common/io.ts
+++ b/src/common/io.ts
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import Arweave from 'arweave';
+
 import { ARIO_TESTNET_PROCESS_ID } from '../constants.js';
 import {
   AoArNSNameDataWithName,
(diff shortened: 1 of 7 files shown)


src/common/io.ts
```diff
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import Arweave from 'arweave';
```
fedellen, line 16: always find myself crossing my fingers when I import this now 😶‍🌫️

src/utils/arweave.ts
```diff
@@ -54,3 +57,66 @@ export const paginationParamsToTags = <T>(
...
+    query: `
+      query {
+        transactions(
+          tags: [
+            { name: "From-Process", values: ["${processId}"] }
+            { name: "Action", values: ["Epoch-Distribution-Notice"] }
+            { name: "Epoch-Index", values: ["${epochIndex}"] }
+            { name: "Data-Protocol", values: ["ao"] }
+          ],
+          owners: ["fcoN_xJeisVsPXA-trzVAuIiqO3ydLQxM-L4XbrQKzY"],
```
fedellen, line 108: nit: bring out to a constant or as an authorities input

fedellen (approved):

src/utils/arweave.ts
dtfiedler: will leave as is for now, we don't have any known trusted authorities. don't want to leave room for potential test addresses that return unexpected data. can revisit post AO mainnet

PR #338, opened 2025-01-13
refactor: catch verbose AOS errors including and parse intended error message PE-7417

fedellen:
image

### DIFF
--- a/src/cli/cli.ts
+++ b/src/cli/cli.ts
@@ -258,7 +258,11 @@ makeCommand({
   description: 'Get observations for an epoch',
   options: epochOptions,
   action: (o) =>
-    readARIOFromOptions(o).getObservations(epochInputFromOptions(o)),
+    readARIOFromOptions(o)
+      .getObservations(epochInputFromOptions(o))
+      .then(
+        (result) => result ?? { message: 'No observations found for epoch' },
+      ),
 });
 
 makeCommand({
@@ -266,7 +270,11 @@ makeCommand({
   description: 'Get distributions for an epoch',
   options: epochOptions,
   action: (o) =>
-    readARIOFromOptions(o).getDistributions(epochInputFromOptions(o)),
+    readARIOFromOptions(o)
+      .getDistributions(epochInputFromOptions(o))
+      .then(
+        (result) => result ?? { message: 'No distributions found for epoch' },
+      ),
 });
 
 makeCommand({
--- a/src/cli/cli.ts
+++ b/src/cli/cli.ts
@@ -18,7 +18,7 @@
 // eslint-disable-next-line header/header -- This is a CLI file
 import { program } from 'commander';
 
-import { spawnANT } from '../node/index.js';
+import { AOProcess, AoMessageResult, spawnANT } from '../node/index.js';
 import { mARIOToken } from '../types/token.js';
 import { version } from '../version.js';
 import {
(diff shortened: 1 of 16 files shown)


src/types/io.ts
```diff
@@ -668,6 +668,7 @@ export interface AoARIOWrite extends AoARIORead {
     params: AoRedelegateStakeParams,
     options?: WriteOptions,
   ): Promise<AoMessageResult>;
+  writeAction(options?: WriteOptions): Promise<AoMessageResult>;
```
fedellen, line 671: this offers some utility here, but we could instead just make an  `AoProcess` in CLI layer and call `send`

fedellen: 17136e4

src/cli/cli.ts
```diff
@@ -987,6 +988,21 @@ makeCommand<
...
+  action: async (options) => {
+    const process = new AOProcess({
+      processId: requiredProcessIdFromOptions(options),
+    });
+    return process.send<AoMessageResult>({
+      tags: writeActionTagsFromOptions(options).tags ?? [],
+      signer: requiredAoSignerFromOptions(options),
+    });
+  },
+});
```
fedellen, line 1004: adds some more utility to the CLI, being able to run custom write actions like 

```sh
node lib/esm/cli/cli.js write-action --process-id KF4wSJa5no2yyDjrDE3hxd5sJoGhLh0UEj8a6Rz3aX8 --wallet-file ../path/to/wallet.json --tags Action Register Registration-Type stake
```

fedellen: nit: name could be more aligned with SDK and require `action`, like `send-action` with `--action` option. same utility could be made for `read-action`

fedellen: fwiw ao.link uses `read` and `write`. that might be more user friendly than `send` -- and both methods will actually "send" and trigger some computation

src/common/contracts/ao-process.ts
```diff
@@ -234,15 +234,21 @@ function errorMessageFromOutput(output: {
   Messages?: { Tags?: { name: string; value: string }[] }[];
 }): string | undefined {
   const errorData = output.Error;
-  if (errorData !== undefined) {
-    // TODO: Could clean this one up too, current error is verbose, but not always deterministic for parsing
-    // Throw the whole raw error if AO process level error
-    return errorData;
+
+  if (errorData?.includes('handling message with Action')) {
+    // Response includes CU AO Error Variant:
```
atticusofsparta, line 239: Shouldnt the CU be spitting this back on result.Error?

src/common/contracts/ao-process.ts
```diff
@@ -234,15 +234,21 @@ function errorMessageFromOutput(output: {
...
+    // "\u001b[31mError\u001b[90m handling message with Action = Register\u001b[0m\n\u001b[32m[string \".handlers\"]:723: [string \"aos\"]:761: Recipient not found for wallet address\u001b[0m\n\n\u001b[90mstack traceback:\n\t[string \".process\"]:871: in function '.process.handle'\u001b[0m\n\n\u001b[31merror:\n\u001b[0m[string \".handlers\"]:723: [string \"aos\"]:761: Recipient not found for wallet address
+
+    const errorsOnLine = errorData.split('\n')[1].split(':');
```
arielmelendez, line 242: I'd suggest being a bit more defensive on the array indexing in this area.

dtfiedler:
just a nit on commit structure - these aren't necessarily refactors. we're implementing improvements of known TODOs, so they are more like fix if they address the todos, feats if they provided additional utility (generally)

dtfiedler (dismissed):

PR #339, opened 2025-01-14
fix(spawn): spawn ANTs with a custom ANT module instead of aos module

atticusofsparta:


dtfiedler (approved):
👏 - we don't have e2e tests for this as ao-localnet is half-baked. but can you confirm this works locally?

atticusofsparta:
@dtfiedler did a local install and it worked great, slightly faster registration too. Me likey.

arielmelendez (approved):

PR #343, opened 2025-01-14
fix(pagination): allow nested keys in sortBy pagination params utility type PE-7428

fedellen:
image

### DIFF
--- a/src/cli/utils.ts
+++ b/src/cli/utils.ts
@@ -302,7 +302,8 @@ export function paginationParamsFromOptions<O extends PaginationCLIOptions>(
   return {
     cursor,
     limit: numberLimit,
-    sortBy,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    sortBy: sortBy as any,
     sortOrder,
   };
 }
--- a/src/cli/utils.ts
+++ b/src/cli/utils.ts
@@ -40,6 +40,7 @@ import {
   FundFrom,
   Logger,
   PaginationParams,
+  SortBy,
   SpawnANTState,
   WriteOptions,
   createAoSigner,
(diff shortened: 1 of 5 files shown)


src/types/io.ts
```diff
@@ -31,11 +31,20 @@ import {
...
 // Pagination
+type NestedKeys<T> = T extends object
+  ? T extends readonly unknown[] // Detect arrays precisely
+    ? never // Exclude arrays
+    : {
+        [K in keyof T & string]: T[K] extends object
+          ? `${K}.${NestedKeys<T[K]>}` | K
+          : K;
+      }[keyof T & string]
+  : never;
```
dtfiedler, line 42: nice

src/types/io.ts
```diff
@@ -31,11 +31,20 @@ import {
...
 
 export type PaginationParams<T = Record<string, never>> = {
   cursor?: string;
   limit?: number;
-  sortBy?: keyof T extends never ? string : keyof T; // default to string if T is empty
+  sortBy?: T extends Record<string, never> ? string : NestedKeys<T>;
```
dtfiedler, line 47: can you add an e2e test for this one with weights?

src/cli/utils.ts
```diff
@@ -302,7 +302,8 @@ export function paginationParamsFromOptions<O extends PaginationCLIOptions>(
   return {
     cursor,
     limit: numberLimit,
-    sortBy,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    sortBy: sortBy as any,
```
dtfiedler, line 306: nit: does `unknown` work better here? otherwise you can provide another type to the generic and do a SortBy<T> on it.

tests/e2e/esm/index.test.ts
```diff
@@ -256,799 +256,823 @@ describe('e2e esm tests', async () => {
...
-
-    it('should be able to get gateway delegates', async () => {
-      const delegates = await ario.getGatewayDelegates({
-        address: 'QGWqtJdLLgm2ehFWiiPzMaoFLD50CnGuzZIPEdoDRGQ',
-        limit: 1,
-        sortBy: 'startTimestamp',
+    it('should be able to get a page of gateways sorted by nested key like `weights.compositeWeight`', async () => {
+      const gateways = await ario.getGateways({
+        limit: 3,
+        sortBy: 'weights.compositeWeight',
```
dtfiedler, line 262: 👏

dtfiedler (dismissed):

fedellen:
got an issue here in the tests, don't merge yet -- will come back to this after some airdrop eventing cycles

src/cli/utils.ts
fedellen: got another generic here for types c29bc32

we're not yet asserting the types though. would need a constant for each set of valid sort bys or derive them from an instantiated object of each

dtfiedler (approved):

PR #345, opened 2025-01-15
fix(types): simplify types for init functions, cleanup contructors

dtfiedler:
I introduced a bug when adding `arweave` as an optional type in https://github.com/ar-io/ar-io-sdk/pull/337. It's easy to overlook what our overloaded `init` functions expect. This hopefully introduces new types that can be extended if/when necessary. Additionally, i've added e2e tests validating the instance types returned for all major classes. We can look to leverage unit tests more to avoid the entire docker setup when doing these simple type validations/checks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
   "dependencies": {
     "@dha-team/arbundles": "^1.0.1",
     "@permaweb/aoconnect": "^0.0.57",
-    "arweave": "1.14.4",
+    "arweave": "1.15.5",
     "axios": "1.7.9",
     "axios-retry": "^4.3.0",
     "commander": "^12.1.0",
--- a/tests/e2e/cjs/index.test.ts
+++ b/tests/e2e/cjs/index.test.ts
@@ -36,6 +36,7 @@ const signers = [
   createAoSigner(new ArweaveSigner(testWallet)),
 ];
 const processId = process.env.ARIO_PROCESS_ID || arioDevnetProcessId;
+const arweave = Arweave.init({});
 describe('e2e cjs tests', async () => {
   describe('ARIO client works ', async () => {
     it('should be able to instantiate ARIO with default process', async () => {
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -126,7 +126,7 @@
   "dependencies": {
     "@dha-team/arbundles": "^1.0.1",
     "@permaweb/aoconnect": "^0.0.57",
-    "arweave": "1.14.4",
+    "arweave": "1.15.5",
```
dtfiedler, line 129: this always makes me nervous. will do some validations across our clients (observer, ar-io-node, network-portal, arns.app) to validate if it causes any issues

tests/e2e/cjs/index.test.ts
```diff
@@ -34,22 +35,47 @@ const signers = [
...
 
+    it.skip('should be able to instantiate ARIO with a process and arweave', async () => {
+      const ario = ARIO.init({
+        process: new AOProcess({
+          processId,
+        }),
+        arweave,
+      });
+      assert(ario instanceof ARIOReadable);
+    });
```
fedellen, line 64: is this meant to be removed or unskipped?

tests/e2e/esm/index.test.ts
```diff
@@ -680,18 +718,20 @@ describe('e2e esm tests', async () => {
...
+        const vault = await ario.getVault({
+          address: vaults[0].address,
+          vaultId: vaults[0].vaultId,
+        });
+        assert.deepEqual(vault, {
+          balance: 1,
+          startTimestamp: 1729962428678,
+          endTimestamp: 1731172028678,
+        });
+      }
```
fedellen, line 733: this doesn't seem stable. that vault will end at 1731172028678. maybe just assert the types?

dtfiedler: yes - good catch

package.json
fedellen: minor version bump -- "Drop Node 16 Support". that semantic versioning 😂

tests/e2e/esm/index.test.ts
fedellen: hmm, `1731172028678` is actually in `GMT: Saturday, November 9, 2024 5:07:08.678 PM`

I'm guessing this is just getting no vaults atm

fedellen (dismissed):
looks good besides the test nits -- can iterate

fedellen (approved):

PR #348, opened 2025-01-15
fix(error handling): trim escape codes from thrown error PE-7417

fedellen:


dtfiedler (dismissed):

src/common/contracts/ao-process.ts
```diff
@@ -253,9 +253,24 @@ function errorMessageFromOutput(output: {
...
+  return error
+    .replace(
+      new RegExp(
+        `${ESC}[\\[\\]()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]`,
+        'g',
+      ),
+      '',
+    )
+    .trim();
+}
```
dtfiedler, line 276: nit: take one of the error messages we get and add it as a unit test

dtfiedler:
```
ao-cu-1  | [string ".handlers"]:723: [string "aos"]:976: Recipient already exists
ao-cu-1  | 
ao-cu-1  | stack traceback:
ao-cu-1  |      [string ".process"]:871: in function '.process.handle'
ao-cu-1  | 
```

tests/e2e/esm/index.test.ts
```diff
@@ -1108,4 +1109,60 @@ describe('e2e esm tests', async () => {
       }
     });
   });
+
+  // Skipping due to inconsistent results with spun up CU
```
fedellen, line 1113: these work when I hit cu.ardrive or use AO infra. but using our local CU this causes a false negative. we have this covered in unit test though

tests/e2e/esm/index.test.ts
```diff
@@ -1108,4 +1109,60 @@ describe('e2e esm tests', async () => {
...
+        })
+        .catch((e) => e);
+
+      assert.ok(error);
+      assert.ok(error instanceof WriteInteractionError);
+      assert.equal(
+        error.message.split('(')[0].trim(),
+        'Recipient not found for wallet address',
+      );
+    });
```
dtfiedler, line 1140: these tests are actually writing message to our airdrop process - we likely don't want to do that

dtfiedler (approved):

PR #366, opened 2025-01-28
fix(ant): add setUndername and setBasename apis

dtfiedler:
This is for better DX when manipulating records on ANTs


fedellen (approved):

atticusofsparta (approved):

atticusofsparta:
Looks good. I do wonder if a unified api on certain actions would be a good idea.

Like an ArNSUser class that initializes with configs for ario, ant, and ant registry processes, then can expose some functions, like setBaseName and setPrimaryName that compose some actions to create workflows, like how with setBaseName it is retrieving the arns record to interact with the ant.

PR #367, opened 2025-01-29
fix(PE-7040): updated weight names

dtfiedler:
Depends on https://github.com/ar-io/ar-io-network-process/pull/341

### DIFF
--- a/src/types/io.ts
+++ b/src/types/io.ts
@@ -242,8 +242,12 @@ export type AoWeightedObserver = {
 export type AoGatewayWeights = {
   stakeWeight: number;
   tenureWeight: number;
+  // @deprecated - use `gatewayPerformanceRatio` instead
   gatewayRewardRatioWeight: number;
+  // @deprecated - use `observerPerformanceRatio` instead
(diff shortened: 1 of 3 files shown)


src/types/io.ts
```diff
@@ -241,8 +241,12 @@ export type AoWeightedObserver = {
 export type AoGatewayWeights = {
   stakeWeight: number;
   tenureWeight: number;
+  // @deprecated - use `gatewayPerformanceRatio` instead
   gatewayRewardRatioWeight: number;
+  // @deprecated - use `observerPerformanceRatio` instead
   observerRewardRatioWeight: number;
+  gatewayPerformanceRatio: number;
+  observerPerformanceRatio: number;
```
fedellen, line 249: in the method, we could duplicated this deprecated value to prevent any backwards compatiblity issues. at least until release 4.0.0

dtfiedler: the contract will continue to provide these values for now - then we can prep the update pre-mainnet and likely release a v4 release with all the final outputs

fedellen (approved):

PR #371, opened 2025-01-29
feat(ant apis): add commands for new methods

fedellen:


src/types/ant.ts
```diff
@@ -202,109 +203,49 @@ export interface AoANTRead {
...
-    options?: WriteOptions,
-  ): Promise<AoMessageResult>;
+  transfer: WriteAction<{ target: WalletAddress }>;
+  addController: WriteAction<{ controller: WalletAddress }>;
+  removeController: WriteAction<{ controller: WalletAddress }>;
+  /** @deprecated Use setUndernameRecord instead for undernames, and setBaseNameRecord instead for the top level name (e.g. "@") */
+  setRecord: WriteAction<AoANTSetUndernameRecordParams>;
+  removeRecord: WriteAction<{ undername: string }>;
+  setBaseNameRecord: WriteAction<AoANTSetBaseNameRecordParams>;
+  setUndernameRecord: WriteAction<AoANTSetUndernameRecordParams>;
```
fedellen, line 213: I found that these new methods did not have `WriteOptions` copy pasted into the parameters. to prevent this in the future I created a new `WriteAction` utility type that will apply the options and the common result. fwiw its only been applied on the ANT Type to consolidate the diff

src/common/ant.ts
```diff
@@ -82,6 +85,8 @@ export class AoANTReadable implements AoANTRead {
     } else {
       throw new InvalidContractConfigurationError();
     }
+
+    this.processId = this.process.processId;
```
fedellen, line 89: applied on this class to easily access the processId from instantiated ANT

atticusofsparta (approved):

src/types/ant.ts
fedellen: also just noticed that it wasn't on the implementation, added here : df830b6

dtfiedler (approved):
thanks - still haven't remembered to update my workflow to include updating the CLI!

PR #381, opened 2025-02-04
feat(mainnet): use mainnet process ID by default

fedellen:
BREAKING CHANGE: SDK and CLI will not longer go to AR.IO Testnet by default

### DIFF
--- a/src/cli/options.ts
+++ b/src/cli/options.ts
@@ -27,8 +27,13 @@ export const optionMap = {
     alias: '--private-key <key>',
     description: 'Stringified private key to use with the action',
   },
-  dev: {
-    alias: '--dev',
+  testnet: {
+    alias: '--testnet',
+    description: 'Run against the AR.IO testnet process',
+    type: 'boolean',
+  },
+  devnet: {
+    alias: '--dev, --devnet',
     description: 'Run against the AR.IO devnet process',
     type: 'boolean',
   },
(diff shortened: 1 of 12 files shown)


src/cli/options.ts
```diff
@@ -27,8 +27,13 @@ export const optionMap = {
...
   },
-  dev: {
-    alias: '--dev',
+  testnet: {
+    alias: '--testnet',
+    description: 'Run against the AR.IO testnet process',
+    type: 'boolean',
+  },
+  devnet: {
+    alias: '--dev, --devnet',
```
fedellen, line 36: allows `--dev` to continue to work, adds `--devnet` and `--testnet`

src/constants.ts
```diff
@@ -37,6 +37,8 @@ export const arioDevnetProcessId = ARIO_DEVNET_PROCESS_ID;
 export const ARIO_TESTNET_PROCESS_ID =
   'agYcCFJtrMG6cqMuZfskIkFTGvUPddICmtQSBIoPdiA';
 
+export const ARIO_MAINNET_PROCESS_ID = 'placeholder'; // TODO: update when mainnet process ID is available
```
fedellen, line 40: insert ID here
```suggestion
export const ARIO_MAINNET_PROCESS_ID = 'mainnet_process_id';
```

dtfiedler:
@fedellen let's prepare this on a `release/v4.0.0` branch and we will add other changes

PR #386, opened 2025-02-05
fix(evolve): use fetch for data instead of arweave

atticusofsparta:


src/utils/ao.ts
```diff
@@ -127,10 +127,14 @@ export async function evolveANT({
...
-    string: true,
-  })) as string;
+  // We do not use arweave to get the data because it may throw on l2 tx data
+  const {
+    api: { host, port, protocol },
+  } = arweave.getConfig();
+
+  const luaString = await fetch(
+    `${protocol}://${host}:${port}/${luaCodeTxId}`,
+  ).then((res) => res.text());
```
fedellen, line 137: as a result of this, can we remove `arweave` as a dependency to this function?

atticusofsparta: we probably can, and switch the dependency to be just a gateway config, but I didn't want to break the api, and we still need to know what gateway to get the data from. So figure it can stay for the moment.

fedellen (approved):

PR #387, opened 2025-02-05
fix(PE-7391): add localAO class for testing

atticusofsparta:
Implements a local AoClient class that satisfies our usage in AOProcess (dryrun, message, and result apis)

Could be fleshed out further in the future.

Operation:
- Maintains an in-memory property of the process memory
- Maintains a results cache
- Maintains and increments a txid-like nonce
- Dryrun does not affect nonce or memory or results cache
- Message increments the nonce, and sets the result to the results cache, return the nonce (as the equivalent of the txid)
- Result consumes the nonce and returns the results for the nonce from the results cache.

`dryrun` `message` and `result` accepts the same parameters as their sister functions from `AoClient`, with some exceptions

`message` and `dryrun` accept handler ENV overrides to pass to AoLoader
`result` does not care about `process` id property from parameters

It may be worth passing in a results cache that IS indexed by process ID for testing purposes for each of fetching results.


package.json
```diff
@@ -125,6 +125,7 @@
   },
   "dependencies": {
     "@dha-team/arbundles": "^1.0.1",
+    "@permaweb/ao-loader": "^0.0.46",
```
dtfiedler, line 128: make this a dev dep

src/common/contracts/ao-process.ts
```diff
@@ -23,6 +24,143 @@ import { version } from '../../version.js';
...
+  inputEncoding: 'JSON-1',
+  outputEncoding: 'JSON-1',
+  memoryLimit: '1073741824', // 1 GiB in bytes
+  computeLimit: (9e12).toString(),
+  extensions: [],
+};
+
+export type HandleFunction = Awaited<ReturnType<typeof AoLoader>>;
+
+export class LocalAO implements Partial<AoClient> {
```
dtfiedler, line 58: put all this in tests, don't want to export it from the SDK nor the related dependencies

atticusofsparta: Oh - I was actually thinking this might be useful for testing on the arns portal. But can figure that out later, will move.

dtfiedler: we can harden the pattern/class here and then potentially move it to a separate testing library

tests/integration/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "type": "module",
+  "scripts": {
+    "postinstall": "yarn link @ar.io/sdk && yarn update-aos-ant",
+    "update-aos-ant": "node update-aos-ant-wasm.mjs",
+    "test": "node --import=../../register.mjs --trace-warnings --test index.test.ts && echo '\nNode ESM Integration ✅'"
+  },
+  "dependencies": {
+    "@ar.io/sdk": "*",
+    "@permaweb/aoconnect": "^0.0.59",
+    "arweave": "^1.15.5"
```
dtfiedler, line 14: these can all be dev deps with this structure, no?

tests/integration/update-aos-ant-wasm.mjs
```diff
@@ -0,0 +1,35 @@
...
+  );
+
+  for (const file of aosAntFiles) {
+    if (!file.includes(AOS_MODULE_ID)) {
+      fs.unlinkSync(path.join(fixturesDir, file));
+    }
+  }
+
+  if (!aosAntFiles.some((file) => file.includes(AOS_MODULE_ID))) {
+    const res = await fetch(`https://arweave.net/${AOS_MODULE_ID}`).then((r) =>
```
fedellen, line 24: nit: inject gateway into the script via env var or similar and default to arweave.net

tests/integration/utils.ts
```diff
@@ -0,0 +1,169 @@
...
+  From: STUB_ADDRESS,
+  Timestamp: Date.now(),
+  // for msg.reply
+  Reference: '1',
+  Data: '1234',
+};
+
+export type HandleFunction = Awaited<ReturnType<typeof AoLoader>>;
+
+export class LocalAO implements Partial<AoClient> {
```
dtfiedler, line 57: nit: add some documentation for this class and how it works/what it does

fedellen (dismissed):
nice pattern, lets get ripping on some write tests 🥂

tests/integration/index.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  async function createLocalANT() {
+    return new AOProcess({
+      processId: 'ant-'.padEnd(43, '1'),
+      ao: (await LocalAO.init({
+        wasmModule: TEST_AOS_ANT_WASM,
+        aoLoaderOptions: AO_LOADER_OPTIONS,
+        handlerEnv: AO_LOADER_HANDLER_ENV,
+      })) as any,
+    });
+  }
```
dtfiedler, line 33: nit: move this into utils and import in the tests

tests/integration/index.test.ts
```diff
@@ -0,0 +1,74 @@
...
+});
+
+const testWallet = JSON.parse(testWalletJSON);
+const signers = [
+  new ArweaveSigner(testWallet),
+  createAoSigner(new ArweaveSigner(testWallet)),
+] as const;
+
+describe('integration esm tests', async () => {
+  async function createLocalANT() {
```
dtfiedler, line 24: is it possible to make this `createLocalProcess` and specify which wasm to use?

tests/integration/utils.ts
```diff
@@ -0,0 +1,169 @@
...
+import path from 'node:path';
+
+const __dirname = path.dirname(new URL(import.meta.url).pathname);
+export const TEST_AOS_ANT_WASM = fs.readFileSync(
+  path.join(__dirname, 'fixtures', `aos-ant-${AOS_MODULE_ID}.wasm`),
+);
+
+export const STUB_PROCESS_ID = 'process-id-'.padEnd(43, '1');
+export const STUB_ADDRESS = 'arweave-address-'.padEnd(43, '1');
+export const STUB_ETH_ADDRESS = '0xFCAd0B19bB29D4674531d6f115237E16AfCE377c';
```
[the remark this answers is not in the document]
atticusofsparta: [5c7b0d0](https://github.com/ar-io/ar-io-sdk/pull/387/commits/5c7b0d0f2fcd67c001c85fa6bd19d180d8c0a48d)

tests/integration/package.json
atticusofsparta: [129ba8d](https://github.com/ar-io/ar-io-sdk/pull/387/commits/129ba8d470f411282778ae8acac043716297d45b)

tests/integration/update-aos-ant-wasm.mjs
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/387/commits/01b1f96936d7654e6dcee3f5bb576b7c67e6846c

tests/integration/utils.ts
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/387/commits/3a583d7de2f539c0e93d8a13a1112b1d28d8086e

tests/integration/test-wallet.json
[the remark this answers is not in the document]
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/387/commits/3a583d7de2f539c0e93d8a13a1112b1d28d8086e

tests/integration/index.test.ts
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/387/commits/3a583d7de2f539c0e93d8a13a1112b1d28d8086e

tests/integration/index.test.ts
atticusofsparta: [7b64273](https://github.com/ar-io/ar-io-sdk/pull/387/commits/7b64273c070d95bb14e372f8ee1e813b09180793)

.github/workflows/build.yml
```diff
@@ -36,7 +36,7 @@ jobs:
   test:
     strategy:
       matrix:
-        suite: [unit, web, esm, cjs]
+        suite: [unit, web, esm, cjs, integration]
```
dtfiedler, line 39: fine to keep - but if we see issues with setup/tests we may want to remove as we continue to harden

tests/integration/utils.ts
```diff
@@ -0,0 +1,220 @@
...
+    processId,
+    ao: (await LocalAO.init({
+      wasmModule,
+      aoLoaderOptions,
+      handlerEnv,
+    })) as any,
+  });
+}
+
+export async function createLocalANT({
```
dtfiedler, line 208: nit - delete this and use the `createLocalProcess` utility

tests/integration/utils.ts
```diff
@@ -0,0 +1,220 @@
...
+ */
+export class LocalAO implements Partial<AoClient> {
+  wasmModule: any;
+  handle: HandleFunction;
+  memory: ArrayBufferLike | null;
+
+  handlerEnv: typeof AO_LOADER_HANDLER_ENV;
+
+  nonce: string;
+  resultsCache: Map<string, Awaited<ReturnType<AoClient['result']>>> =
```
dtfiedler, line 86: i think it's fine, but curious to understand why you want to use a map here vs just "lastResult" or something similar. some concern this could grow unbounded in tests (may be better as an LRU). 

don't change it, just help me understand the thought process

dtfiedler (approved):
some questions, but excited to start to leveraging this and get more coverage. nice work!

tests/integration/utils.ts
atticusofsparta: Agree with the LRU - also probably OK with lastResult, the thinking for me here was Map made it simple to imitate and integrate with how we current use AoClient in the AOProcess class.

Was also thinking of the other AoClient APIs like `results` - which i didn't implement because we don't currently use it - which returns paginated responses for all transactions for a process. So Map seemed like a good candidate.

dtfiedler (approved):

PR #390, opened 2025-02-09
fix(sdk): compute epoch index based on 1-index

dtfiedler:


dtfiedler:
We're not going to do this.

PR #406, opened 2025-02-14
fix(PE-7718): improved retry logic

kunstmusik:
For AoProcess, only treat exceptions from ao.dryrun(), ao.message(), and ao.result() as retry-able errors. All other errors are treated as valid results from the process.

### DIFF
--- a/src/common/contracts/ao-process.ts
+++ b/src/common/contracts/ao-process.ts
@@ -20,6 +20,7 @@ import {
   AoClient,
   AoSigner,
   DryRunResult,
+  MessageResult,
 } from '../../types/index.js';
 import { getRandomText } from '../../utils/base64.js';
 import { errorMessageFromOutput } from '../../utils/index.js';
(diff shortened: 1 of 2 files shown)


dtfiedler:
nit: this commit would be considered a fix vs. a feat

kunstmusik:
Ugh, just saw the comment about fix vs feat.  I may need to redo the changes to change the commit messages.

src/common/contracts/ao-process.ts
```diff
@@ -60,72 +66,83 @@ export class AOProcess implements AOContract {
...
-        // return undefined if no data is returned
-        if (this.isMessageDataEmpty(messageData)) {
-          return undefined as K;
-        }
-
-        const response: K = safeDecode<K>(messageData);
-        return response;
-      } catch (error: any) {
+        result = await this.ao.dryrun(dryRunInput);
+        // break on successful return of result
```
atticusofsparta, line 88: We still want to check the result for the error and construct an error to throw it i think.
```
 const error = errorMessageFromOutput(result);
        if (error !== undefined) {
          throw new Error(error);
        }
```

atticusofsparta (approved):

dtfiedler (approved):

PR #431, opened 2025-03-03
fix(schemas): update ant schema to accept eth support

atticusofsparta:


src/types/ant.ts
```diff
@@ -64,10 +78,10 @@ export type SortedANTRecords = Record<string, SortedANTRecord>;
 
 export const AntRecordsSchema = z.record(z.string(), AntRecordSchema);
 export const AntControllersSchema = z.array(
-  ArweaveTxIdSchema.describe('Controller address'),
+  AOAddressSchema.describe('Controller address'),
```
dtfiedler, line 81: ...what if we relaxed this to just be any string?

atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/431/commits/090c799a9b366960a6f29d6d9c89d78c086ce444

dtfiedler (approved):

PR #432, opened 2025-03-04
feat(PE-7756): ant registry versions

atticusofsparta:


src/common/ant-registry.ts
```diff
@@ -85,6 +86,14 @@ export class AoANTRegistryReadable implements AoANTRegistryRead {
       ],
     });
   }
+
+  async getVersions(): Promise<
+    Record<string, { moduleId: string; luaSourceId?: string; notes: string }>
```
dtfiedler, line 91: jc - why is luaSourceId optional

src/common/ant-registry.ts
```diff
@@ -85,6 +86,14 @@ export class AoANTRegistryReadable implements AoANTRegistryRead {
       ],
     });
   }
+
+  async getVersions(): Promise<
```
dtfiedler, line 90: nit: make this getANTVersions

dtfiedler: also, add `getLatestANTVersion`

dtfiedler: i know we _store_ the version in formation on the registry, but wondering if this is better as a static attribute on ANT class with it's own interface defined, separate from the ANT Registry. The ANT registry is just one _source_ of where these versions could be stored


```typescript
export class ANT {
    static versions: ANTVersions = new ANTProcessVersions();
    
    
    // existing constructor and all the other ANT APIs 
 }
 
interface ANTVersions {
    getVersions()
    getLatest()
 }
 
// any process that holds ANT version state, defaults to ANT registry
 export class ANTProcessVersions implements ANTVersions {
      private process: AOProcess;
       constructor({ process = new AOProcess({ processId: ANT_REGISTRY_ID }): AOProcess } ){
             this.process = process
       }
       
       getVersions(){
              // ao send message to ANT process
              return this.process.send({ // tags for fetching versions })
       }
       
       getLatest() {
            // get versions, return latest
       }
 }
```

fedellen (approved):

src/common/contracts/ant-versions.ts
```diff
@@ -0,0 +1,127 @@
...
+    this.signer = createAoSigner(signer);
+  }
+  async addVersion(params: {
+    version: string;
+    moduleId: string;
+    luaSourceId?: string;
+    notes?: string;
+  }) {
+    return this.process.send({
+      tags: pruneTags([
```
dtfiedler: i feel like these should all be required

dtfiedler (approved):
overall looks good, one question if `notes` and `luaSourceCodeId` should be required

src/common/contracts/ant-versions.ts
atticusofsparta: The contract takes them as optional so i left them optional here

PR #436, opened 2025-03-12
fix(PE-7794): log errors more verbosely on read fails

atticusofsparta:


src/common/contracts/ao-process.ts
```diff
@@ -160,6 +166,8 @@ export class AOProcess implements AOContract {
     let attempts = 0;
     let messageId: string | undefined;
     let result: MessageResult | undefined = undefined;
+    // anchor is a random text produce non-deterministic messages IDs when deterministic signers are provided (ETH)
+    const anchor = getRandomText(32);
```
atticusofsparta, line 170: applying anchor here makes this api a bit opinionated - this could be done in the signer thats passed in of non deterministic messages are desired. That being said, moved it up out of the loop so that atleast in this send call we expect ONE message only.

dtfiedler (approved):

PR #438, opened 2025-03-20
fix(gql): add fallback to go to CU for epoch distribution data

dtfiedler:
If the epoch distribution message does not get cranked, we need to fallback to to the CU to with any tick messages that couldve created it and parse the results of those messages. This is a temporary patch, while we work with FWD to resolve

### DIFF
--- a/src/utils/arweave.ts
+++ b/src/utils/arweave.ts
@@ -16,7 +16,7 @@
 import Arweave from 'arweave';
 
 import { ARIO_MAINNET_PROCESS_ID, ARWEAVE_TX_REGEX } from '../constants.js';
-import { BlockHeight } from '../types/common.js';
+import { AoClient, BlockHeight } from '../types/common.js';
 import {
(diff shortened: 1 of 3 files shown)


src/utils/arweave.ts
```diff
@@ -155,6 +243,37 @@ export const epochDistributionNoticeGqlQuery = ({
   });
   return gqlQuery;
 };
+// fallback query if the distribution notice does not get cranked
+export const epochDistributionNoticeGqlQueryFallback = ({
+  processId = ARIO_MAINNET_PROCESS_ID,
+  owners = ['OAb-n-ZugyN598kZNpfOy0ACelGVmwCQ0kYbgNGDUK8'], // ar.io team wallet ticks once a day
```
dtfiedler, line 249: this wallet is scheduled to tick once a day right at the distribution timestamp, if any other wallet ticks before it, then the query won't work...temporary solution

PR #441, opened 2025-03-25
Fix AOProcess import and resolve TypeScript errors

internettrashh:
the example for using custom config here : https://github.com/ar-io/ar-io-sdk?tab=readme-ov-file#configuration
would not have worked since the import was missing.

What was fixed:
Added missing AOProcess import from @ar.io/sdk
Fixed the TypeScript error: Cannot find name 'AoProcess'


dtfiedler:
Thanks for flagging @internettrashh!

dtfiedler (dismissed):

PR #458, opened 2025-04-15
PE-7751 config changes

atticusofsparta:
Modifies the turbo arns class to have authenticated and unauthenticated versions, updates signer to be more restricted types based on the message signing we have implmented.

Would like to add some stronger typing on the ARIO writable class to indicate on apis with `fundFrom: "turbo"` whether or not they can be used with turbo base on the class configuration.

### DIFF
--- a/src/common/io.ts
+++ b/src/common/io.ts
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Signer } from '@dha-team/arbundles';
 import Arweave from 'arweave';
 
 import { ARIO_MAINNET_PROCESS_ID } from '../constants.js';
@@ -93,7 +92,11 @@ import {
 import { defaultArweave } from './arweave.js';
 import { AOProcess } from './contracts/ao-process.js';
 import { InvalidContractConfigurationError } from './error.js';
-import { TurboArNSPaymentProvider } from './turbo.js';
+import {
+  TurboArNSPaymentProviderAuthenticated,
+  TurboArNSPaymentProviderUnauthenticated,
+  isTurboArNSSigner,
+} from './turbo.js';
 
 type ARIOConfigNoSigner = OptionalPaymentUrl<
   OptionalArweave<ProcessConfiguration>
@@ -101,7 +104,24 @@ type ARIOConfigNoSigner = OptionalPaymentUrl<
 type ARIOConfigWithSigner = WithSigner<
   OptionalPaymentUrl<OptionalArweave<ProcessConfiguration>>
 >;
-type ARIOConfig = ARIOConfigNoSigner | ARIOConfigWithSigner;
+
+type WithAuthenticatedPaymentProvider<T> = T & {
+  paymentProvider: TurboArNSPaymentProviderAuthenticated;
+};
+
+type WithUnauthenticatedPaymentProvider<T> = T & {
+  paymentProvider: TurboArNSPaymentProviderUnauthenticated;
+};
+
+type ARIOConfigWithSignerAndAuthenticatedPaymentProvider =
+  WithAuthenticatedPaymentProvider<ARIOConfigWithSigner>;
+type ARIOConfigWithSignerAndUnauthenticatedPaymentProvider =
+  WithUnauthenticatedPaymentProvider<ARIOConfigWithSigner>;
+type ARIOConfig =
+  | ARIOConfigNoSigner
+  | ARIOConfigWithSigner
+  | ARIOConfigWithSignerAndAuthenticatedPaymentProvider
+  | ARIOConfigWithSignerAndUnauthenticatedPaymentProvider;
 
 export class ARIO {
   // Overload: No arguments -> returns AoARIORead
@@ -126,7 +146,7 @@ export class ARIOReadable implements AoARIORead {
   protected process: AOProcess;
   protected epochSettings: AoEpochSettings | undefined;
   protected arweave: Arweave;
-  protected paymentProvider: TurboArNSPaymentProvider; // TODO: this could be an array/map of payment providers
+  protected paymentProvider: TurboArNSPaymentProviderUnauthenticated; // TODO: this could be an array/map of payment providers
 
   constructor(config?: ARIOConfigNoSigner) {
     this.arweave = config?.arweave ?? defaultArweave;
@@ -143,7 +163,7 @@ export class ARIOReadable implements AoARIORead {
     } else {
       throw new InvalidContractConfigurationError();
     }
-    this.paymentProvider = new TurboArNSPaymentProvider({
+    this.paymentProvider = new TurboArNSPaymentProviderUnauthenticated({
       paymentUrl: config?.paymentUrl,
     });
   }
@@ -860,7 +880,9 @@ export class ARIOReadable implements AoARIORead {
 export class ARIOWriteable extends ARIOReadable implements AoARIOWrite {
   protected declare process: AOProcess;
   private signer: AoSigner;
-  protected paymentProvider: TurboArNSPaymentProvider;
+  protected paymentProvider:
+    | TurboArNSPaymentProviderAuthenticated
+    | TurboArNSPaymentProviderUnauthenticated;
 
   constructor({ signer, paymentUrl, ...config }: ARIOConfigWithSigner) {
     if (config === undefined) {
@@ -873,10 +895,14 @@ export class ARIOWriteable extends ARIOReadable implements AoARIOWrite {
       super(config);
     }
     this.signer = createAoSigner(signer);
-    this.paymentProvider = new TurboArNSPaymentProvider({
-      signer: signer as Signer,
-      paymentUrl,
-    });
+    this.paymentProvider = isTurboArNSSigner(signer)
+      ? new TurboArNSPaymentProviderAuthenticated({
+          signer,
+          paymentUrl,
+        })
+      : new TurboArNSPaymentProviderUnauthenticated({
+          paymentUrl,
+        });
   }
 
   async transfer(
@@ -1274,7 +1300,10 @@ export class ARIOWriteable extends ARIOReadable implements AoARIOWrite {
     params: AoBuyRecordParams,
     options?: WriteOptions,
   ): Promise<AoMessageResult> {
-    if (params.fundFrom === 'turbo') {
+    if (
+      this.paymentProvider instanceof TurboArNSPaymentProviderAuthenticated &&
+      params.fundFrom === 'turbo'
+    ) {
       return this.paymentProvider.initiateArNSPurchase({
         intent: 'Buy-Name',
         ...params,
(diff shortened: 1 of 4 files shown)


src/common/turbo.ts
```diff
@@ -225,3 +279,26 @@ export class TurboArNSPaymentProvider implements ArNSPaymentProvider {
...
+    'getActivePublicKey' in signer
+  );
+}
+
+export function isTurboArNSSigner(signer: unknown): signer is TurboArNSSigner {
+  const isWanderWallet = isWanderArweaveBrowserSigner(signer);
+  const isSigner =
+    signer instanceof EthereumSigner ||
+    signer instanceof InjectedEthereumSigner ||
+    signer instanceof ArweaveSigner;
```
fedellen, line 302: add ArconnectSigner

src/common/turbo.ts
```diff
@@ -47,31 +58,57 @@ export async function signedRequestHeadersFromSigner({
...
-    // case SignatureConfig.SOLANA:
-    // case SignatureConfig.ED25519:
-    default:
-      throw new Error(
-        `Unsupported signer type for signing requests: ${signer.signatureType}`,
-      );
-  }
+  let signature: string | undefined = undefined;
+  let publicKey: string | undefined = undefined;
+
```
fedellen, line 63:
```suggestion
if (signer instanceof ArconnectSigner) signer = ArconnectSigner.signer;
```

src/common/io.ts
```diff
@@ -950,10 +956,14 @@ export class ARIOWriteable extends ARIOReadable implements AoARIOWrite {
...
-      paymentUrl,
-    });
+    this.paymentProvider = isTurboArNSSigner(signer)
+      ? new TurboArNSPaymentProviderAuthenticated({
+          signer,
+          paymentUrl,
+        })
+      : new TurboArNSPaymentProviderUnauthenticated({
+          paymentUrl,
+        });
```
dtfiedler, line 966: i'd prefer to just have a factory utility or class that returns the signer vs. the ternary here

src/common/turbo.ts
atticusofsparta: Not a fan of overriding arguments, but this accomplishes the same thing: [35ae6aa](https://github.com/ar-io/ar-io-sdk/pull/458/commits/35ae6aa2f2e7a5429c35df9a4a4cabbbb499fed3)

src/common/turbo.ts
atticusofsparta: [35ae6aa](https://github.com/ar-io/ar-io-sdk/pull/458/commits/35ae6aa2f2e7a5429c35df9a4a4cabbbb499fed3)

src/common/io.ts
```diff
@@ -1445,7 +1464,10 @@ export class ARIOWriteable extends ARIOReadable implements AoARIOWrite {
     params: AoIncreaseUndernameLimitParams,
     options?: WriteOptions,
   ): Promise<AoMessageResult> {
-    if (params.fundFrom === 'turbo') {
+    if (
+      this.paymentProvider instanceof TurboArNSPaymentProviderAuthenticated &&
```
dtfiedler, line 1468: replace paymentProviders to be an Record<'turbo', PaymentProvider> and then modify this logic to do `this.paymentProvider[fundFrom].initiatiteArNSPurchase()`

src/common/io.ts
```diff
@@ -203,7 +207,7 @@ export class ARIOReadable implements AoARIORead {
   public readonly process: AOProcess;
   protected epochSettings: AoEpochSettings | undefined;
   protected arweave: Arweave;
-  protected paymentProvider: TurboArNSPaymentProvider; // TODO: this could be an array/map of payment providers
+  protected paymentProvider: TurboArNSPaymentProviderUnauthenticated; // TODO: this could be an array/map of payment providers
```
dtfiedler, line 210: see other comment - modify this to be Record<string, PaymentProvider>; and access it via `fundFrom`

src/common/turbo.ts
dtfiedler: does it need to be an instance of one of these, or just have a `sign()` function?

src/types/common.ts
```diff
@@ -39,7 +45,13 @@ export type OptionalArweave<T = NonNullable<unknown>> = {
...
   paymentUrl?: string;
 } & T;
-export type TurboArNSSigner = Signer;
+// TODO: TurboArNSSigner could be simply `Signer` but we need to implement each message signing method for signing headers.
+export type TurboArNSSigner =
+  | EthereumSigner
+  | InjectedEthereumSigner
+  | ArweaveSigner
+  | ArconnectSigner
+  | Window['arweaveWallet'];
```
dtfiedler, line 54: I'm considered we are not sufficiently abstracting signers. Ideally any `Signer` used satisfies the necesssary interface we determine based on the integration points with AO and Turbo. It already seems confusing to me you could have a `TurboArNSSigner` which is not satisfied by an `AoSigner` or vice versa

dtfiedler: how do we get unit tests for each signer type? what kind of life that would be?

src/common/turbo.ts
```diff
@@ -47,31 +59,66 @@ export async function signedRequestHeadersFromSigner({
...
+    signature = toB64Url(
+      Buffer.from(
+        await signer.signMessage(Uint8Array.from(Buffer.from(nonce))),
+      ),
+    );
+  } else if (
+    signer instanceof ArweaveSigner ||
+    signer instanceof EthereumSigner ||
+    signer instanceof InjectedEthereumSigner
+  ) {
```
dtfiedler, line 88: isn't this the as `isTurboArNSSigner`?

src/common/turbo.ts
```diff
@@ -82,36 +129,41 @@ export type ArNSPurchaseReceipt = AoTokenCostParams & {
   createdDate: string;
 };
 
-export interface ArNSPaymentProvider {
+// Define separate provider interfaces
+export interface ArNSUnauthenticatedPaymentProvider {
```
dtfiedler, line 133:
```suggestion
export interface ArNSPaymentProvider {
```

src/common/turbo.ts
```diff
@@ -82,36 +129,41 @@ export type ArNSPurchaseReceipt = AoTokenCostParams & {
...
     mARIO: mARIOToken;
   }>;
+}
+
+export interface ArNSAuthenticatedPaymentProvider
+  extends ArNSUnauthenticatedPaymentProvider {
   initiateArNSPurchase(
     params: AoTokenCostParams & { processId?: TransactionId },
-    options: WriteOptions,
+    options: WriteOptions, // Note: WriteOptions might not be necessary here if unused
```
dtfiedler, line 147: remove it

src/common/turbo.ts
atticusofsparta: No - isTurboArNSSigner includes browser signer which doesnt implement the Signer interface from arbundles.

src/common/io.ts
atticusofsparta: [0527b83](https://github.com/ar-io/ar-io-sdk/pull/458/commits/0527b8357fe47f1dee8831832b43a520555a5770)

src/common/io.ts
atticusofsparta: so this would translate to `this.paymentProviders['stakes' | 'balances' | 'any' | 'turbo'] ?

dtfiedler: you'd keep the `(if params.fundFrom == 'turbo')` but remove `this.paymentProvider instanceof TurboArNSPaymentProviderAuthenticated`

src/types/common.ts
atticusofsparta: I agree - the issue we are facing here is we made the decision to not use data items for auth and instead sign messages - the tradeoff there is we need to access specific methods, like signMessage on arweaveWallet api, to be able to execute that. It also means we need to implement the api on each specific signer type. Might be able to be a bit looser here and use ContractSigner to a degree - can't use AoSigner because that one is _specifically_ for signing data items, but I think we can abstract it to Signer and Window['arweaveWallet'] atleast.

src/common/turbo.ts
atticusofsparta: So we would have ArNSPaymentProvider and ArNSAuthenticatedPaymentProvider?

src/types/common.ts
atticusofsparta: I think I _can_ use Signer but its a bit wierd to allow all SIgners when in the headers signing we throw if its not an arweave or eth signature type

fedellen (dismissed):

src/types/common.ts
atticusofsparta: node signers are easy to unit test - the web ones a little less so. Injected provider can probably be unit tested with an ethers non web provider, but arconnect gets tricky, needs a mock.

src/common/turbo.ts
atticusofsparta: [336d52f](https://github.com/ar-io/ar-io-sdk/pull/458/commits/336d52f03cd5ed7cb49173e26f4e3838a9b7cf91)

src/common/turbo.ts
```diff
@@ -90,28 +138,72 @@ export interface ArNSPaymentProvider {
     winc: string;
     mARIO: mARIOToken;
   }>;
+}
+
+export interface ArNSAuthenticatedPaymentProvider extends ArNSPaymentProvider {
```
dtfiedler, line 143:
```suggestion
export interface ArNSSignedPaymentProvider extends ArNSPaymentProvider {
```

src/common/turbo.ts
```diff
@@ -90,28 +138,72 @@ export interface ArNSPaymentProvider {
...
+    options: WriteOptions, // Note: WriteOptions might not be necessary here if unused
   ): Promise<AoMessageResult<ArNSPurchaseReceipt>>;
 }
 
-export class TurboArNSPaymentProvider implements ArNSPaymentProvider {
-  private readonly paymentUrl: string;
-  private readonly axios: AxiosInstance;
-  private readonly logger: ILogger;
-  private readonly signer?: TurboArNSSigner;
+export class TurboArNSPaymentFactory {
```
dtfiedler, line 150: TurboPaymentProvider
```suggestion
export class TurboPaymentProvider {
```

src/common/turbo.ts
```diff
@@ -90,28 +138,72 @@ export interface ArNSPaymentProvider {
...
+    return new TurboArNSPaymentProviderUnauthenticated({
+      paymentUrl,
+      axios,
+      logger,
+    });
+  }
+}
+
+// Base class for unauthenticated operations
+export class TurboArNSPaymentProviderUnauthenticated
```
dtfiedler, line 192:
```suggestion
export class TurboArNSPaymentProvider
```

src/common/turbo.ts
```diff
@@ -164,6 +256,22 @@ export class TurboArNSPaymentProvider implements ArNSPaymentProvider {
     const { winc } = await this.getArNSPriceDetails(params);
     return +winc;
   }
+}
+
+// Class for authenticated operations, extending the base class
+export class TurboArNSPaymentProviderAuthenticated
```
dtfiedler, line 262:
```suggestion
export class TurboArNSSignedPaymentProvider
```

src/types/common.ts
dtfiedler: do you intend on adding tests?

src/common/io.ts
atticusofsparta: this is a bit of a wierd state because we still need to ensure the payment provider is authenticated - due to the inconsistent signers, it might not be. Is the intention here to allow for easily selecting other payment providers in the future?

fedellen (approved):

PR #462, opened 2025-04-21
feat(wayfinder): initial wayfinder interfaces & http client wrapper for ar:// support

dtfiedler:
A new generic utility function, `createWayfinderClient`, wraps any compatible HTTP client with support for the ar:// protocol, and selects the target gateway based on the provided strategy. With it, you can:
- get the redirectUrl - identifies and replaces ar:// with selected gateway url
- request the data using the provided http client (will eventually allow us to support data verification)

The top-level `Wayfinder` class can be used to construct the Wayfinder client and strategy. 
Below are some examples that the current implementation supports with various TODOs for future features

```typescript
const wayfinder = new Wayfinder({
   // pick one of the top 3 gateways based on to their totalDelegatedStake
   router: new PriorityGatewayRouter({ 
     gatewaysProvider: ARIOGateawaysProvider({ 
        ario: ARIO.mainnet(), 
     });
    sortBy: 'totalDelegatedStake', 
    sortOrder: 'desc', 
     limit: 3
   })
   // use axios HTTP client
    httpClient: axios,
   // TODO: add default verification settings here
});

// get the redirect url using the provided strategy
const redirectUrl = await wayfinder.resolveUrl({ originalUrl: 'ar://example' });

// fetch the data from the url - using the wrapped axios http client (optionally support data verification here)
const response: AxiosResponse = await wayfinder.request('ar://example', {
   headers: {
      'Content-Type': 'html/text',
   },
   // TODO: add per request verification settings here (e.g. disable verification for this request)
});


// supports use the fetch client
const wayfinderWithFetch = new Wayfinder({
   // use a random gateway
   router: new RandomGatewayRouter({ gatewaysProvider: ARIOGateawaysProvider({ ario: ARIO.mainnet() ) }),
   httpClient: fetch,
});

// since we're using fetch, the response for .request is Response
const response: Response = await wayfinderWithFetch.request('ar://example')


// it also supports more advanced HTTP clients like got
const wayfinderWithGot = new Wayfinder({
    // use a random gateway
   router: new RandomGatewayRouter({ gatewaysProvider: ARIOGateawaysProvider({ ario: ARIO.mainnet() ) }),
   httpClient: got,
});

// returns a Response stream object from got (see test for example)
const stream = await wayfinder.request.stream('ar://example');
```


src/common/wayfinder.ts
```diff
@@ -0,0 +1,255 @@
...
+// TODO: introduce wayfinder http wrapper class/interface so we can support a variety of http clients to proxy requests through
+export type AnyFunction = (...args: any[]) => any;
+export type WayfinderHttpClient<T extends AnyFunction> = T;
+export const createWayfinderHttpClient = <T extends AnyFunction>({
+  httpClient,
+}: {
+  httpClient: T; // generic http client parameters
+}): WayfinderHttpClient<T> => {
+  return new Proxy(httpClient, {
+    async apply(
```
dtfiedler, line 164: we could potentially just provide the dependencies to `Wayfinder` here and expose those function to get a wayfinder HTTP client, which allows you to provide any supported `ar://` route directly to it, and exposes `wayfinder.getRedirectUrl` on the fetch client

dtfiedler: moved to `createWayfinderClient` which accepts a `resolveUrl` function that is used to get the redirect URL. this gives us flexibility to introduce new clients (e.g. rpc) and allows us to add data verification

src/common/wayfinder/routers/random.ts
```diff
@@ -0,0 +1,55 @@
...
+import { AoARIORead } from '../../../types/io.js';
+import { WayfinderRouter } from '../../../types/wayfinder.js';
+import { randomInt } from '../wayfinder.js';
+
+export class RandomGatewayRouter implements WayfinderRouter {
+  public readonly name = 'random';
+  private ario: AoARIORead;
+  private blocklist: string[];
+  constructor({
+    ario,
```
dtfiedler, line 25: it may be easier to make this just anything that supports `getGateways` interface - allowing us to do robust caching of the full list of gateways

package.json
```diff
@@ -110,6 +110,7 @@
     "eslint-plugin-n": "^16.0.1",
     "eslint-plugin-prettier": "^5.0.0",
     "eslint-plugin-promise": "^6.1.1",
+    "got": "^12.4.7",
```
dtfiedler, line 113: added as a dev dependency for testing

package.json
```diff
@@ -92,7 +92,7 @@
     "@semantic-release/git": "^10.0.1",
     "@semantic-release/npm": "^11.0.3",
     "@trivago/prettier-plugin-sort-imports": "^4.2.0",
-    "@types/node": "^20.12.12",
+    "@types/node": "^22.14.1",
```
arielmelendez, line 95: Should this coincide with a .nvmrc change?

arielmelendez: OH duh I skipped right over it. Nice.

src/common/io.ts
```diff
@@ -940,7 +940,6 @@ export class ARIOReadable implements AoARIORead {
 }
 
 export class ARIOWriteable extends ARIOReadable implements AoARIOWrite {
-  public declare readonly process: AOProcess;
```
arielmelendez, line 943: Remind me - why was this excised?

arielmelendez (approved):
Looks great for initial PoC!

PR #466, opened 2025-04-29
feat(wayfinder): add initial data verification interfaces and classes

dtfiedler:
The implementation supports trusted gateway hash providers, and digest verification using hte fetch http client. More to do to support data root verification and other http clients but this is a start.


examples/vite/src/App.tsx
dtfiedler, line 1: any easy web based client example using wayfinder to route/verify tx ids

src/common/wayfinder/verification/trusted-gateway.ts
```diff
@@ -0,0 +1,98 @@
...
+ * limitations under the License.
+ */
+import { DataHashProvider, DataVerifier } from '../../../types/wayfinder.js';
+
+const arioGatewayHeaders = {
+  digest: 'x-ar-io-digest',
+  verified: 'x-ar-io-verified',
+};
+
+export class TrustedGatewayHashProvider implements DataHashProvider {
```
djwhitt, line 23: Nit: consider making this plural - `TrustedGatewaysHashProvider`.

src/types/wayfinder.ts
```diff
@@ -13,7 +13,39 @@
...
+}
+
+export interface DataHashProvider {
+  /**
+   * Returns a hash for the provided txId. Depending on the implementation, the hash could be a data root, a digest, or some other hash of the data.
+   *
+   * @param txId - The txId of the data
+   * @returns the hash of the data
+   */
+  getHash: ({ txId }: { txId: string }) => Promise<string>;
```
djwhitt, line 50: I wonder if it's worth including the hash type in the return value. Not sure, but worth considering.

dtfiedler: I had the same thought, and considering making it a required attribute on the interface instead of on this method.

package.json
```diff
@@ -72,7 +72,7 @@
     "test:cjs": "yarn build:cjs && yarn link && cd ./tests/e2e/cjs && yarn && yarn test",
     "test:esm": "yarn build:esm && yarn link && cd ./tests/e2e/esm && yarn && yarn test",
     "test:web": "yarn build:esm && yarn link && cd ./tests/e2e/web && yarn && yarn test",
-    "test:unit": "node --import=./register.mjs --test --test-reporter=spec --enable-source-maps --trace-warnings 'src/**/*.test.ts'",
+    "test:unit": "node --import=./register.mjs --test --test-reporter=spec --enable-source-maps --trace-warnings 'src/**/wayfinder.test.ts'",
```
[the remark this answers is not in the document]
dtfiedler: will revert once i finish up this PR

src/common/wayfinder/verification/data-root-verifier.ts
```diff
@@ -0,0 +1,111 @@
...
+import { Readable } from 'node:stream';
+
+import { DataRootProvider, DataVerifier } from '../../../types/wayfinder.js';
+import {
+  hashBufferToB64Url,
+  hashReadableStreamToB64Url,
+  hashReadableToB64Url,
+} from '../../../utils/hash.js';
+
+export async function convertBufferToDataRoot({
```
djwhitt, line 32: Calculating data roots this way is fine, but it may not reproduce the original data root. We could also implement another data root verifier that pulls all the chunks, checks that they reproduce the expected data root, and then compares the concatenated chunk data to the original data retrieved. That would take a while, bit it should be able to verify any L1 data where we can find the chunks.

src/common/wayfinder/verification/trusted-gateway.ts
dtfiedler: [d9ca652](https://github.com/ar-io/ar-io-sdk/pull/466/commits/d9ca652d1ed0b4b2ee553f7f148d6738ca0dbbf3)

PR #470, opened 2025-05-15
feat(arns): add ArNSNameResovler interface and API 

dtfiedler:
This interface can be impemented in various ways. For now just adds resolution API directly via CUs on ARIO class. Wayfinder can also implement it, using router strategies to resolve arns names against ar-io resolver endpoints.

CLI:
```bash
❯ node lib/esm/cli/cli.js resolve-arns-name --name dtf
{
  "name": "dtf",
  "owner": "ZjmB2vEUlHlJ7-rgJkYP09N5IzLPhJyStVrK5u9dDEo",
  "txId": "Zkoo3vgTq-f0wBlM-cOYhVy3dxVMajr7l-dEBFRYos4",
  "ttlSeconds": 3600,
  "recordIndex": 0,
  "processId": "Rfm9Ob2M-7FRKdi4azJVU4JLpUa6bjM1Ud3KpwkepOE",
  "undernameLimit": 10,
  "type": "permabuy"
}
```

### DIFF
--- a/CLI.md
+++ b/CLI.md
@@ -91,24 +91,24 @@ Options:
   -h, --help                              display help for command
 
 Commands:
+  # ARIO Network
+
+  # Getters
   info [options]                          Get network info
   token-supply [options]                  Get the total token supply
+  balance [options]                       Get the balance of an address
   get-registration-fees [options]         Get registration fees
   get-demand-factor [options]             Get demand factor
   get-demand-factor-settings [options]    Get current settings for demand factor
   get-epoch-settings [options]            Get current settings for epochs
   get-gateway [options]                   Get the gateway of an address
-  list-gateways [options]                 List the gateways of the network
-  list-all-delegates [options]            List all paginated delegates from all gateways
   get-gateway-delegates [options]         Get the delegates of a gateway
+  get-gateway-vaults [options]            Get the vaults of a gateway
   get-delegations [options]               Get all stake delegated to gateways from this address
   get-allowed-delegates [options]         Get the allow list of a gateway delegate
   get-arns-record [options]               Get an ArNS record by name
-  list-arns-records [options]             List all ArNS records
   get-arns-reserved-name [options]        Get a reserved ArNS name
-  list-arns-reserved-names [options]      Get all reserved ArNS names
   get-arns-returned-name [options]        Get an ArNS returned name by name
-  list-arns-returned-names [options]      Get all ArNS recently returned names
   get-epoch [options]                     Get epoch data
   get-current-epoch [options]             Get current epoch data
   get-prescribed-observers [options]      Get prescribed observers for an epoch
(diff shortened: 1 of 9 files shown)


tests/e2e/esm/index.test.ts
```diff
@@ -1162,6 +1162,30 @@ describe('e2e esm tests', async () => {
...
+      });
+      assert.ok(arnsName);
+      assert.equal(typeof arnsName.owner, 'string');
+      assert.equal(typeof arnsName.name, 'string');
+      assert.equal(typeof arnsName.processId, 'string');
+      assert.equal(typeof arnsName.ttlSeconds, 'number');
+      assert.equal(typeof arnsName.txId, 'string');
+    });
+
+    it('should be able to resolve an undername', async () => {
```
dtfiedler, line 1177: nit: add tests for error scenarios (e.g. name and undername not found)

dtfiedler:
nit: add docs for using this API

CLI.md
```diff
@@ -91,24 +91,24 @@ Options:
   -h, --help                              display help for command
 
 Commands:
+  # ARIO Network
```
fedellen, line 94: previously I was just updating this list from the help command 
image

src/types/io.ts
```diff
@@ -565,7 +565,23 @@ export type DemandFactorSettings = {
...
-export interface AoARIORead {
+// simple interface to allow multiple implementations of ArNSNameResolver
+export type ArNSNameResolutionData = {
+  name: string;
+  owner?: string; // could be unowned
+  txId: string;
+  type: 'lease' | 'permabuy';
+  processId: string;
+  ttlSeconds: number;
+  recordIndex?: number; // TODO: the SDK should always provide a priority index, even if the ANT does not have a priority set
```
fedellen, line 576: should this be non-nullable as the retrun from SDK? the comment seems to suggest so

also why the change from `priority` to `recordIndex`?

src/cli/cli.ts
```diff
@@ -620,6 +628,8 @@ makeCommand({
   action: requestPrimaryNameCLICommand,
 });
 
+// ANTS
```
fedellen, line 631: 🐜 🐜 🐜

fedellen (approved):

src/types/io.ts
dtfiedler: I added other TODOs to enforce priority is set on all ant `getRecords` requests. I'd like to separate that change and test in isolation. After that I'll come back and remove these `todos`

this API essentially normalizes the arns name data on the ARIO contract and the record data from the ANT. 


> also why the change from priority to recordIndex?

priority is a overloaded, but i agree it should be consistent. will modify

CLI.md
dtfiedler: i'll organize these to match - the readme

dtfiedler: [4d410f0](https://github.com/ar-io/ar-io-sdk/pull/470/commits/4d410f096d8dedcd2e3c7cd275486f81216120f9)

PR #481, opened 2025-05-22
feat(ants): add support for fetching ant state from hyperbeam nodes

dtfiedler:
This is an early implementation that needs some tuning, but the intent is with newer ANTs publishing patch messages of their state on changes, we can now start to lean in to using hyperbeam for cached state and lower the need for CUs. We may want to modify how we are verifying HyperBeam compatability vs. the HEAD check against a known hyperbeam node.'

### DIFF
--- a/src/common/ant.ts
+++ b/src/common/ant.ts
@@ -79,6 +79,7 @@ export class AoANTReadable implements AoANTRead {
   protected process: AOProcess;
   public readonly processId: string;
   private strict: boolean;
+  private checkHyperBeamPromise: Promise<boolean> | undefined;
 
   constructor(config: ANTConfigOptionalStrict) {
     this.strict = config.strict || false;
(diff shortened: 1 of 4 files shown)


src/common/ant.ts
```diff
@@ -93,11 +94,81 @@ export class AoANTReadable implements AoANTRead {
     }
 
     this.processId = this.process.processId;
+    this.checkHyperBeamPromise = this.checkHyperBeamCompatibility();
+  }
+
+  private async checkHyperBeamCompatibility(): Promise<boolean> {
+    const res = await fetch(
+      `https://permanode.xyz/${this.processId}~process@1.0/now/cache`,
```
dtfiedler, line 102: TODO: add support for custom hyperbeam URL via constructor

src/common/ant.ts
```diff
@@ -93,11 +94,81 @@ export class AoANTReadable implements AoANTRead {
...
+          (acc, [key, record]) => {
+            acc[key] = {
+              transactionId: record.transactionid,
+              ttlSeconds: record.ttlseconds,
+            };
+            return acc;
+          },
+          {} as Record<string, AoANTRecord>,
+        ),
+        Balances: body.balances,
```
dtfiedler, line 165: confirming with FWD that the json@1.0 device makes all keys lowercase and there is no way to avoid it - consider pulling this conversion to a utility function vs. baked into this implementation

src/common/ant.ts
```diff
@@ -119,6 +190,21 @@ export class AoANTReadable implements AoANTRead {
...
+      return {
+        Name: state.Name,
+        Ticker: state.Ticker,
+        Description: state.Description,
+        Keywords: state.Keywords,
+        Denomination: state.Denomination.toString(),
+        Owner: state.Owner,
+        Logo: state.Logo,
+        'Total-Supply': state.TotalSupply.toString(),
+        Handlers: [], // TODO: support for handler in patched state
```
dtfiedler, line 204: @atticusofsparta - do we use Handlers anymore?

src/common/ant.ts
atticusofsparta: well thats mighty annoying. Is there a reason why it does that?

src/common/ant.ts
```diff
@@ -93,11 +100,50 @@ export class AoANTReadable implements AoANTRead {
...
+      {
+        method: 'HEAD',
+      },
+    );
+
+    if (res.ok) {
+      return true;
+    }
+
+    return false;
```
atticusofsparta, line 119: Rather than doing:

```
if (await this.checkHyperBeamPromise) ...
```
we can do:
```
if (await this.checkHyperBeamCompatibility())
```

Which itself can check this.checkHyperBeamPromise internally and also handle retries. If this throws for some reason or has a bad status, this will allow us to revalidate the compatibility easily with little changes in the future.

atticusofsparta: We should also add some error handling I think here, along with logging of the status / errors

src/common/ant.ts
atticusofsparta: We do but shouldnt. Its been on my list to make a feature flag map and list it along with every version mapped to the module ID in the ant registry. Needing the handlers means that we need to do another call to check api support for enabling features in the portal. I would rather load the handlers based on the version and then be able to have the ANT class know whether or not to throw based on its handlers state.

src/common/ant.ts
```diff
@@ -259,6 +355,29 @@ export class AoANTReadable implements AoANTRead {
...
+          },
+        },
+      );
+      if (!res.ok) {
+        throw new Error('Failed to fetch ant balances');
+      }
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { device, ...balances } = (await res.json()) as {
+        device: string;
+        [key: string]: any;
```
atticusofsparta, line 376: should be string: number

Then wont have to cast the return i think

src/common/ant.ts
atticusofsparta: Agreed, we don't need to add retries now, but we can memoize _in the function_ instead of checking the promise, this way if we do eventually change the pattern we can simply modify the logic in the function and not have to worry about changing the locations using this.promiseResult.

Its not a huge change to the current pattern:
```suggestion
  private async checkHyperBeamCompatibility(): Promise<boolean> {
    if (this.checkHyperBeamPromise !== undefined) {
      return this.checkHyperBeamPromise;
    }

    const res = await fetch(
      `https://permanode.xyz/${this.processId}~process@1.0/now/cache`,
      {
        method: 'HEAD',
      },
    );

    let isHyperBeamCompatible = false;

    if (res.ok) {
      isHyperBeamCompatible = true;
    }

    this.checkHyperBeamPromise = Promise.resolve(isHyperBeamCompatible);

    return isHyperBeamCompatible;
  
```

atticusofsparta: constructor call would remain, and then using it would look like:
```ts
await this.checkHyperBeamCompatibility()
```

and it would use the memoized value

dtfiedler: yeah - something like that, just confirming the memoize is called and hyperbeam used. we can add with some more testing.

atticusofsparta (approved):

PR #483, opened 2025-05-22
feat(wayfinder): create composite verification strategy

dtfiedler:
Pipes streams to multiple verification strategies for parallel verification. Errors on the first failure

### DIFF
--- a/src/common/wayfinder/verification/strategies/composite-verifier.ts
+++ b/src/common/wayfinder/verification/strategies/composite-verifier.ts
@@ -1,0 +1,140 @@
+/**
+ * Copyright (C) 2022-2024 Permanent Data Solutions, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { PassThrough, Readable } from 'node:stream';
+
+import { DataVerificationStrategy } from '../../../../types/wayfinder.js';
+
+/**
+ * A verification strategy that composes multiple verification strategies together.
+ * This allows for more comprehensive verification by applying multiple verification techniques.
+ * All verifications are run in parallel, and the process stops on the first failure.
+ */
+export class CompositeVerificationStrategy implements DataVerificationStrategy {
+  private readonly verificationStrategies: DataVerificationStrategy[];
+
+  /**
+   * Creates a new CompositeVerificationStrategy
+   *
+   * @param verificationStrategies - An array of verification strategies to apply
+   */
+  constructor({
+    verificationStrategies,
+  }: {
+    verificationStrategies: DataVerificationStrategy[];
+  }) {
+    if (
+      verificationStrategies === undefined ||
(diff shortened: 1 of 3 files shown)


src/common/wayfinder/verification/strategies/composite-verifier.ts
```diff
@@ -0,0 +1,140 @@
...
+  }: {
+    data: Buffer | Readable | ReadableStream;
+    txId: string;
+  }): Promise<void> {
+    // for Buffer data, reuse the same buffer for all verifications
+    if (Buffer.isBuffer(data)) {
+      await Promise.all(
+        this.verificationStrategies.map(async (strategy, index) => {
+          try {
+            await strategy.verifyData({ data, txId });
```
arielmelendez, line 67: In thinking about these running in parallel now, and the possibility that one of the verifiers may fail fast, I'm somewhat desiring to see abort controllers passed through to verifyData and shared among implementations...

src/common/wayfinder/verification/strategies/composite-verifier.ts
```diff
@@ -0,0 +1,140 @@
...
+      return;
+    }
+
+    // For Node.js Readable streams
+    if (data instanceof Readable) {
+      // create a pass through stream for each strategy and pipe the data to it
+      const streams = this.verificationStrategies.map(() => new PassThrough());
+
+      for (const stream of streams) {
+        data.pipe(stream);
```
arielmelendez, line 85: I don't think this approach is what you want. From the pipe docs:
```
The readable.pipe() method attaches a [Writable](https://nodejs.org/api/stream.html#class-streamwritable) stream to the readable, causing it to switch automatically into flowing mode and push all of its data to the attached [Writable](https://nodejs.org/api/stream.html#class-streamwritable). The flow of data will be automatically managed so that the destination Writable stream is not overwhelmed by a faster Readable stream.
```
So once you attach the first `stream` receiver, you're going to start exhausting your `data` stream right away. You'd probably fill it to the highWaterMark and then the next stream would fill up to its highWaterMark with the remaining bytes, and so on.

I think what you want is to use our usual pattern of pausing the input (`data`) stream, adding .on('data') handlers to each `stream`, getting a references to your Promise.all promise without awaiting it, calling resume() on the `data` stream, and then awaiting Promise.all for a result.

src/common/wayfinder/verification/strategies/composite-verifier.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  async verifyData({
+    data,
+    txId,
+  }: {
+    data: Buffer | Readable | ReadableStream;
+    txId: string;
+  }): Promise<void> {
+    // for Buffer data, reuse the same buffer for all verifications
+    if (Buffer.isBuffer(data)) {
```
arielmelendez, line 63: You can save yourself some code by just turning all the buffers into Readables and letting it fall into the next case. Up to you though.

arielmelendez: Downside of my suggestion is what I mention in comment down below about the limitation of split streams; your verifiers will only go as fast as the slowest verifier. By using the buffer directly like this, faster verifiers can fail faster than slower verifiers and potentially save time and resources. If you keep this block here you may want to mention that as a reason for doing so.

arielmelendez (changes requested):
There's a Readable splitting issue, I believe. Might be worth figuring out why tests didn't catch it if so.

src/common/wayfinder/verification/strategies/composite-verifier.ts
dtfiedler: we actually can lean more into `AsyncIterable`. Both `Buffer` and `Readable` satisfy that type and `ReadableStreams` are fairly easy to convert to one, allowing us to simplify some of these if blocks. good learnings to take to the turbo-sdk as well.

PR #485, opened 2025-05-28
fix(wayfinder): resolve ARNS names with path components correctly

Bobinstein:
The resolveWayfinderUrl function was incorrectly testing the entire path against the ARNS regex instead of just the name portion. This caused URLs like ar://cookbook_ao/welcome/getting-started.html to fail ARNS validation because the regex was testing cookbook_ao/welcome/getting-started.html instead of just cookbook_ao.

Changes: - Split path into name and remaining components before regex testing - Test only the first part against txIdRegex and arnsRegex - Improve URL construction to properly handle paths with and without additional components - Maintain backward compatibility for existing functionality

Fixes routing for ARNS names that include path components while preserving all existing behavior for transaction IDs and simple ARNS names.

### DIFF
--- a/src/common/wayfinder/wayfinder.test.ts
+++ b/src/common/wayfinder/wayfinder.test.ts
@@ -652,4 +652,333 @@ describe('Wayfinder', () => {
       });
     });
   });
+
+  describe('URL resolution', () => {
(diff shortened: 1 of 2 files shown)


dtfiedler (approved):
looks great - thanks for the tests!

PR #489, opened 2025-06-02
feat(arns): add referer support to purchase APIs for tracking (PE-8088)

Bobinstein:
Adds optional `referer` parameter to all ArNS purchase APIs (buyRecord, upgradeRecord, extendLease, increaseUndernameLimit) to enable tracking of applications that facilitate purchases for honeycomb event pipeline statistics and incentive programs.

The referer parameter is passed through both direct AR.IO process calls (via message tags) and Turbo payment flows (via URL parameters). CLI commands also support the new `--referer` option.

All changes are backwards compatible as the referer parameter is optional.


package.json
```diff
@@ -140,6 +140,9 @@
     "winston": "^3.13.0",
     "zod": "^3.23.8"
   },
+  "resolutions": {
+    "@permaweb/aoconnect": "0.0.57"
+  },
```
dtfiedler, line 145:
```suggestion
```

dtfiedler (changes requested):
Looks good - please update the `README` examples with the opiontal parameter

Bobinstein:
README updated

dtfiedler (commented):
fun fact - `referer` is famously misspelled in http protocol header spec - https://en.wikipedia.org/wiki/HTTP_referer

i made the same mistake when writing the ticket. let's update these all to use `Referrer`

Bobinstein:
all uses of `referer` have been updated to `referrer`

PR #502, opened 2025-06-20
fix(PE-8070): add retries to hyperbeam calls and fallback to legacy net

atticusofsparta:


src/common/ant.ts
```diff
@@ -138,25 +138,46 @@ export class AoANTReadable implements AoANTRead {
...
-      );
-
-      if (!res.ok) {
-        throw new Error('Failed to fetch ant state');
+      try {
+        let res: Response | undefined = undefined;
+
+        let retries = 0;
+        while (retries < 3) {
+          res = await fetch(
```
dtfiedler, line 146: hyperbeam docs have been updated - https://hyperbeam.ar.io/build/pathing-in-hyperbeam.html

need to learn more, but seems if we use `/now` on the initial check, we can use `/compute` for faster retrieval of the cached state to avoid recomputing latest every time. we may also want to consider memoizing the cache locally for a certain amount of time with a TTL to avoid constant calls

atticusofsparta: yeah, seems like /now is counter to using hyperbeam. /compute is the way to go there.

I figure HEAD on now is ok.

For memoizing the cache... watcha think about a singleton in-memory cache that our instances (ARIO, ANTRegistry, and ANTs) can use as the default cache? TTL would still exist there, and protects against users needlessly reinstantiating classes erasing the cached state, while allowing it to be overridden for external providers?

atticusofsparta: Can also be an LRU cache to prevent gigantic memories of state for clients that load a lot of different ants (like ario nodes)

src/common/ant.ts
```diff
@@ -138,25 +138,46 @@ export class AoANTReadable implements AoANTRead {
...
+              },
+            },
+          );
+          if (res.ok) {
+            break;
+          } else {
+            retries++;
+            await new Promise((resolve) =>
+              setTimeout(resolve, 1000 * retries ** 2),
+            );
```
dtfiedler, line 163: nit: should likely retry only on 5xx errors

dtfiedler (approved):

PR #508, opened 2025-06-30
feat(PE-8228): hb ant registry

atticusofsparta:


src/common/ant-registry.ts
```diff
@@ -55,8 +60,10 @@ export class ANTRegistry {
 
 export class AoANTRegistryReadable implements AoANTRegistryRead {
   protected process: AOProcess;
+  private hyperbeamUrl: string | undefined;
+  private checkHyperBeamPromise: Promise<boolean> | undefined;
 
-  constructor(config?: ProcessConfiguration) {
+  constructor(config?: ANTRegistryConfigOptionalStrict) {
```
dtfiedler, line 66: nit: unclear type name `ANTRegistryConfigOptionalStrict`

src/common/ant-registry.ts
```diff
@@ -70,6 +77,41 @@ export class AoANTRegistryReadable implements AoANTRegistryRead {
...
+  private async checkHyperBeamCompatibility(): Promise<boolean> {
+    if (this.hyperbeamUrl === undefined) {
+      return Promise.resolve(false);
+    }
+    if (this.checkHyperBeamPromise !== undefined) {
+      return this.checkHyperBeamPromise;
+    }
+    console.debug('Checking HyperBeam compatibility');
+    this.checkHyperBeamPromise = fetch(
+      `${this.hyperbeamUrl.toString()}${this.process.processId}~process@1.0/now/acl`,
```
dtfiedler, line 101:
```suggestion
      `${this.hyperbeamUrl.toString()}${this.process.processId}~process@1.0/now/acl`,
```
per Tom's rec, shouldn't we be using `/now/cache/acl`

atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/508/commits/873d356739797f861d2d5bd3c042888f52c4b995

src/common/ant-registry.ts
```diff
@@ -29,9 +29,16 @@ import {
...
-type ANTRegistryWithSigner = WithSigner<ProcessConfiguration>;
-type ANTRegistryConfig = ANTRegistryNoSigner | ANTRegistryWithSigner;
+type ANTRegistryConfig = Required<ProcessConfiguration> & {
+  strict?: boolean;
+  hyperbeamUrl?: string;
+};
+
+type ANTRegistryNoSigner = ANTRegistryConfig;
+type ANTRegistryWithSigner = WithSigner<ANTRegistryConfig>;
+type ANTRegistryConfigOptionalSigner =
```
dtfiedler, line 39: nit: call this ANTRegistryConfig and the existing ANTRegistryConfig as ANTRegistryProcessConfig or similar

src/common/ant-registry.ts
```diff
@@ -70,6 +79,41 @@ export class AoANTRegistryReadable implements AoANTRegistryRead {
...
+      this.checkHyperBeamPromise = this.checkHyperBeamCompatibility();
+    }
+  }
+
+  /**
+   * Check if the process is HyperBeam compatible. If so, we'll use the HyperBeam node to fetch the state.
+   *
+   * @returns {Promise<boolean>} True if the process is HyperBeam compatible, false otherwise.
+   */
+  private async checkHyperBeamCompatibility(): Promise<boolean> {
```
dtfiedler, line 94: not: put in utility and use it various places, including ANTs

src/common/ant-registry.ts
atticusofsparta: https://github.com/ar-io/ar-io-sdk/pull/508/commits/0970a50d8f9baefe6224324251f28dc359cb7856

src/common/ant-registry.ts
atticusofsparta: This feels a bit funky because its a util thats modifying a property on the class (checkHyperBeamPromise)

How about we start a HyperBEAM class that stores the promise there, can be built with the logger, url, processId, and its just this method for the moment? Then the classes like ANT and ANTRegistry can be built with that in the same concept of the AOProcess class.

Later we can add read api to it as well, eg for compute vs cache with pathing and device configuration.

Thoughts?

dtfiedler (approved):

PR #515, opened 2025-07-30
feat(arns): add API to fetch arns names for a given address

dtfiedler:
## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to list all ArNS names associated with a specific address, with support for pagination and registry selection.
  * Introduced a new API method to retrieve paginated ArNS records filtered by wallet address and access control.

* **Bug Fixes**
  * Reduced unnecessary imports in the minimized bundle for improved efficiency.

* **Chores**
  * Updated version to 3.16.1-alpha.1 and upgraded development dependencies.
  * Improved build process to generate and ignore build metadata files.
  * Removed outdated Wayfinder documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1400,6 +1400,49 @@ Available `sortBy` options are any of the keys on the record object, e.g. `name`
 
 </details>
 
+#### `getArNSRecordsForAddress({ address, antRegistryProcessId, cursor, limit, sortBy, sortOrder })`
+
+Retrieves all registered ArNS records of the specified address according to the `ANTRegistry` access control list, paginated and sorted by the specified criteria. The `cursor` used for pagination is the last ArNS name from the previous request.
+
+```typescript
+const ario = ARIO.mainnet();
+const records = await ario.getArNSRecordsForAddress({
+  address: 't4Xr0_J4Iurt7caNST02cMotaz2FIbWQ4Kbj616RHl3',
+  limit: 100,
+  sortBy: 'startTimestamp',
+  sortOrder: 'desc',
+});
+```
+
+Available `sortBy` options are any of the keys on the record object, e.g. `name`, `processId`, `endTimestamp`, `startTimestamp`, `type`, `undernames`.
+
+<details>
+  <summary>Output</summary>
+
+```json
+{
+  "limit": 1,
+  "totalItems": 31,
+  "hasMore": true,
+  "nextCursor": "ardrive",
+  "items": [
+    {
+      "startTimestamp": 1740009600000,
+      "name": "ardrive",
+      "endTimestamp": 1777328018367,
+      "type": "permabuy",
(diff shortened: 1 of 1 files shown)


src/common/io.ts
```diff
@@ -1000,6 +1004,72 @@ export class ARIOReadable implements AoARIORead, ArNSNameResolver {
...
+    PaginationResult<AoArNSNameDataWithName & { role: 'controller' | 'owner' }>
+  > {
+    const { antRegistryId = ANT_REGISTRY_ID, address } = params;
+    // create the ant registry
+    const antRegistry = ANTRegistry.init({
+      process: new AOProcess({
+        ao: this.process.ao,
+        processId: antRegistryId,
+      }),
+    });
```
dtfiedler, line 1032: need to think about this more - if the list ever changes than we break our pagination logic, we likely need to return the processIds in the result, so it can be called recursively to respect pagination.

src/common/io.ts
```diff
@@ -1000,6 +1004,72 @@ export class ARIOReadable implements AoARIORead, ArNSNameResolver {
...
+        processId: allProcessIds,
+      },
+    });
+
+    const namesWithRole: (AoArNSNameDataWithName & {
+      role: 'controller' | 'owner';
+    })[] = namesForPage.items.map((name) => ({
+      ...name,
+      role: Owned.includes(name.processId) ? 'owner' : 'controller',
+    }));
```
atticusofsparta, line 1066: I like the thought, but just want to mention this _might not be accurate_ if the ANT involved isn't sending correct state notices, or more likely, that the MU isn't cranking the notices.

src/common/io.ts
```diff
@@ -1000,6 +1004,72 @@ export class ARIOReadable implements AoARIORead, ArNSNameResolver {
...
+      };
+    }
+
+    const namesForPage = await this.getArNSRecords({
+      ...params,
+      filters: {
+        // TODO: this filters cannot be larger than whatever AO allows for dry-runs, investigate and batch the requests if needed
+        processId: allProcessIds,
+      },
+    });
```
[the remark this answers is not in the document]
atticusofsparta: ftr I tested with 1000 process IDs which equates to 41kb and the tag limit is theoretically 4kb.

src/common/io.ts
```diff
@@ -1000,6 +1004,72 @@ export class ARIOReadable implements AoARIORead, ArNSNameResolver {
...
+    if (allProcessIds.length === 0) {
+      return {
+        items: [],
+        limit: 0,
+        totalItems: 0,
+        sortOrder: 'asc',
+        hasMore: false,
+      };
+    }
+
```
atticusofsparta, line 1052: Erm, shouldn't we be paginating here with processId as the cursor?

PR #520, opened 2025-08-05
fix(deps): reduce included imports in minimized bundle

dtfiedler:


atticusofsparta:
Ideally we have a playwright or simliar tool for a simple html use test i think. But looks good as is.

PR #521, opened 2025-08-05
feat(ant): add fork capability for ANTs

dtfiedler:


src/utils/ao.ts
```diff
@@ -288,7 +293,103 @@ export async function spawnANT({
...
+      ticker: state.Ticker,
+      description: state.Description,
+      keywords: state.Keywords,
+      controllers: state.Controllers,
+      records: state.Records,
+      balances: state.Balances,
+      logo: state.Logo,
+    },
+    onSigningProgress,
+  });
```
atticusofsparta, line 361: To here we could then pass in our `ao` argument as well

```suggestion
  const newProcessId = await spawnANT({
    signer,
    antRegistryId,
    logger,
    ao,
    state: {
      owner: state.Owner,
      name: state.Name,
      ticker: state.Ticker,
      description: state.Description,
      keywords: state.Keywords,
      controllers: state.Controllers,
      records: state.Records,
      balances: state.Balances,
      logo: state.Logo,
    },
    onSigningProgress,
  });
```

src/utils/ao.ts
```diff
@@ -288,7 +293,103 @@ export async function spawnANT({
...
+    throw new Error(
+      `ARNs name (${arnsName}) is not affiliated with the provided ant (${ant.processId})`,
+    );
+  }
+
+  if (nameAfterReassign.processId !== newProcessId) {
+    throw new Error(
+      `ARNs name (${arnsName}) reassignment failed to propagate to the registry. It is still associated with the old ANT (${ant.processId})`,
+    );
+  }
```
atticusofsparta, line 384: Feel like this should all happen in the reassign method

src/utils/ao.ts
```diff
@@ -288,7 +293,103 @@ export async function spawnANT({
...
+      controllers: state.Controllers,
+      records: state.Records,
+      balances: state.Balances,
+      logo: state.Logo,
+    },
+    onSigningProgress,
+  });
+
+  // now send a reassign name notice to the old ant with the new process id
+  const reassignMsgId = await ant.reassignName({
```
atticusofsparta, line 364: Was just going over the upgrade workflow from the ARNS portal and something came up - we should check the min ANT version here and if it is a version that is unrecognized _or_ doesn't support Reassign-Name then we need to eval the source code and _then_ reassign the name to the new ant.

src/utils/ao.ts
```diff
@@ -291,7 +291,59 @@ export async function spawnANT({
   return processId;
 }
 
-// TODO: add a utility for forking an ANT to the latest module that leverages getState and spawnANT
+export async function forkANT({
+  signer,
+  antProcessId,
+  logger = Logger.default,
+  ao,
+  antRegistryId = ANT_REGISTRY_ID,
```
atticusofsparta, line 299: add signing progress and pass to spawnANT

src/utils/ao.ts
```diff
@@ -291,7 +291,59 @@ export async function spawnANT({
...
+    throw new Error(
+      `ANT state (${antProcessId}) is undefined and cannot be upgraded`,
+    );
+  }
+
+  const forkedProcessId = await spawnANT({
+    signer,
+    antRegistryId,
+    logger,
+    state: {
```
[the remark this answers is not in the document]
dtfiedler: this seems fine for now...what are your concerns?

src/common/ant.ts
```diff
@@ -80,13 +80,20 @@ export class ANT {
...
    */
   static spawn = spawnANT;
-  // no signer give read
-  static init(config: ANTConfigNoSigner): AoANTRead;
+  /**
+   * Fork an ANT to a new process.
+   *
+   * @param config
+   */
+  static fork = forkANT;
```
dtfiedler, line 88: i also wonder if we should add this on the processId as a AoWrite which would just call `fork` with itself

dtfiedler: Added

PR #535, opened 2025-08-15
fix(turbo): apply publicKey to signed req headers on node PE-8459

fedellen:
## Summary by CodeRabbit

* Bug Fixes
  * Ensures Arweave-based signers now consistently include public keys in request headers, preventing intermittent authentication/signing failures.

* Improvements
  * Improves reliability and interoperability of signed requests across wallets/providers by using available signer public keys as a fallback for more consistent request handling.

### DIFF
--- a/src/common/turbo.ts
+++ b/src/common/turbo.ts
@@ -86,6 +86,10 @@ export async function signedRequestHeadersFromSigner({
     if ('setPublicKey' in signer && signer['publicKey'] === undefined) {
       await signer.setPublicKey();
     }
+
+    if (signer.publicKey !== undefined) {
+      publicKey = toB64Url(signer.publicKey);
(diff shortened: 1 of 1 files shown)


src/common/turbo.ts
```diff
@@ -98,7 +98,10 @@ export async function signedRequestHeadersFromSigner({
       } else if ('setPublicKey' in signer) {
         await signer.setPublicKey();
         publicKey = toB64Url(signer.publicKey);
+      } else if ((signer as ArweaveSigner).publicKey !== undefined) {
```
dtfiedler, line 101: does this
```suggestion
      } else if ('publicKey' in signer) {
```
Work?

dtfiedler (approved):

PR #541, opened 2025-08-27
feat(PE-8400): add ANT `upgrade` API with callbacks

dtfiedler:


src/common/ant.ts
```diff
@@ -87,6 +89,84 @@ export class ANT {
...
+      logger,
+      antRegistryId,
+      onSigningProgress,
+    });
+
+    // Step 2: Create a writable instance of the original ANT for reassigning names
+    const writableAnt = ANT.init({
+      process: new AOProcess({ processId: antProcessId, ao, logger }),
+      signer,
+    });
```
atticusofsparta, line 144: hb url

src/common/ant.ts
```diff
@@ -87,6 +91,99 @@ export class ANT {
...
+    onSigningProgress?: (
+      name: keyof SpawnAntProgressEvent | 'reassigning-name',
+      payload:
+        | SpawnAntProgressEvent[keyof SpawnAntProgressEvent]
+        | { name: string },
+    ) => void;
+  }): Promise<{
+    forkedProcessId: string;
+    reassignedNames: string[];
+    failedReassignedNames: string[];
```
atticusofsparta, line 134: would be good to make this Map<name, Error> i think.

src/common/ant.ts
```diff
@@ -87,6 +91,99 @@ export class ANT {
...
+    });
+
+    // we could parallelize this, but then signing progress would be harder to track
+    const reassignedNames: string[] = [];
+    const failedReassignedNames: string[] = [];
+    for (const name of names) {
+      try {
+        onSigningProgress?.('reassigning-name', { name });
+
+        await writableAnt.reassignName({
```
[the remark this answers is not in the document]
atticusofsparta: Regarding that we could have an onSigningProgress for cranking, which would looking something like

```typescript
onSigningProgress?.('cranking-reassign', { message: reassignRes.result.Messages[..find reassign message sent to ario process] })
```

Later, could add tooling that tracks the network paths there, eg to the SU, then grabs the message ID, and gets the result to call onSigningProgress with.

src/common/ant.ts
```diff
@@ -87,6 +91,99 @@ export class ANT {
...
+    arioProcessId = ARIO_MAINNET_PROCESS_ID,
+    ao,
+    logger = Logger.default,
+    antRegistryId,
+    skipVersionCheck = false,
+    antProcess = ANT.init({
+      process: new AOProcess({ processId: antProcessId, ao }),
+      signer,
+    }),
+    onSigningProgress,
```
atticusofsparta, line 114: I would recommend we:
- make this a non static method, leveraging the set signer, process id, aoClient, and logger, etc.
- add module ID and tags as options - tags good to track what integration is upgrading, module id for CLI usage and testing / custom versions.
- hyperbeam URL for leveraging cached state

src/common/ant.ts
```diff
@@ -87,6 +91,99 @@ export class ANT {
...
+
+        reassignedNames.push(name);
+      } catch (error) {
+        logger.error(`Failed to reassign name ${name}:`, { error });
+        // Continue with other names rather than failing completely
+        failedReassignedNames.push(name);
+      }
+    }
+
+    return { forkedProcessId, reassignedNames, failedReassignedNames };
```
atticusofsparta, line 184: could be nice to return newAntProcess - essentially clones the current one with the configurations used including signer, ao, etc. 

If remaining with a static function this would use the provided params to construct it, else it would use the configured class properties.

src/common/ant.ts
dtfiedler: there is already a non-static method that calls this method

src/common/ant.ts
```diff
@@ -87,6 +91,99 @@ export class ANT {
...
+    });
+
+    // Step 2: Create a writable instance of the original ANT for reassigning names
+    const writableAnt = ANT.init({
+      process: new AOProcess({ processId: antProcessId, ao, logger }),
+      signer,
+    });
+
+    // we could parallelize this, but then signing progress would be harder to track
+    const reassignedNames: string[] = [];
```
atticusofsparta, line 164: This adds a very specific edge case to the workflow. Im mixed on whether we should have one name or multiple.


On one hand: 

If we fork then reassign 1/2 names, then fail to reassign the second (rate limit, cranking issues) then the other name when upgraded will get a different ANT.

I think we should just have this do one name and not concern ourselves with others, as we cannot guarantee this behaviour (due to infra reliability and limits, especially in batch calls).

Having multiple names on a single arns name is itself an edge case and if clients want to support that, then maybe thats a different workflow. But as is we should just do name -> new process, even if theres multiple connected to the same process.

Its still forking the ANT state, so the owner and records will all be the same, just not centralized to one ANT, and its fixable (via reassign) if the user wants to return their domains to the original configuration.

atticusofsparta (commented):
Would be good to add a cli command here to test, like `ar.io upgrade-domain <name>` which resolve the domain, grabs the process ID, then uses this upgrade flow

I can actually see that being used in a CI workflow in conjunction with permaweb deploy for automatic updates. 🤔

src/common/ant.ts
dtfiedler: > Barring that we can also supply a processId parameter to allow the caller to indicate an existing ANT to reassign to.

This would be just a reassignment, not an upgrade and reassign.

src/common/ant.ts
dtfiedler: [4337a19](https://github.com/ar-io/ar-io-sdk/pull/541/commits/4337a19c85bcf49ca7893af66c0fb108efab9ed9)

atticusofsparta (approved):

PR #563, opened 2025-10-15
fix(PE-8636): add hb client and use in ario client for balance fetching

atticusofsparta:
## Summary by CodeRabbit

* **New Features**
  * Added `--hyperbeam-url` global CLI option to configure custom Hyperbeam node endpoints for routing protocol queries
  * Balance lookups and read operations now automatically utilize Hyperbeam when available, improving query response times

* **Tests**
  * Updated end-to-end test infrastructure to use remote service endpoints

### DIFF
--- a/src/common/hyperbeam/hb.ts
+++ b/src/common/hyperbeam/hb.ts
@@ -20,6 +20,7 @@ export type HBConfig = {
   url: string;
   processId: string;
   logger?: Logger;
+  hbTimeoutMs?: number;
 };
 
 export interface Hyperbeam {
@@ -38,7 +39,7 @@ export interface Hyperbeam {
     path: string;
     json?: boolean;
   }): Promise<T>;
-  checkHyperBeamCompatibility(): Promise<boolean>;
+  checkHyperBeamCompatibility(params?: { minSlot?: number }): Promise<boolean>;
 }
 
 export class HB implements Hyperbeam {
@@ -49,12 +50,13 @@ export class HB implements Hyperbeam {
 
   protected checkHyperBeamPromise: Promise<boolean> | undefined;
   private logger: Logger;
-
+  private hbTimeoutMs: number;
   constructor(config: HBConfig) {
     this.url = config.url;
     this.processId = config.processId;
     this.logger = config.logger ?? Logger.default;
 
+    this.hbTimeoutMs = config.hbTimeoutMs ?? 5000;
     this.isHyperBeamCompatible = undefined;
     this.checkHyperBeamPromise = this.checkHyperBeamCompatibility();
   }
@@ -132,10 +134,20 @@ export class HB implements Hyperbeam {
    *
    * @returns {Promise<boolean>} True if the process is HyperBeam compatible, false otherwise.
    */
-  async checkHyperBeamCompatibility(): Promise<boolean> {
+  async checkHyperBeamCompatibility({
+    minSlot,
+  }: {
+    minSlot?: number;
+  } = {}): Promise<boolean> {
+    // refetch if min slot is provided
+    if (minSlot !== undefined) {
+      this.isHyperBeamCompatible = undefined;
+      this.checkHyperBeamPromise = undefined;
+    }
     if (this.checkHyperBeamPromise !== undefined) {
       return this.checkHyperBeamPromise;
     }
+
     if (this.isHyperBeamCompatible !== undefined) {
       return Promise.resolve(this.isHyperBeamCompatible);
     }
@@ -145,11 +157,21 @@ export class HB implements Hyperbeam {
       `${this.url.toString()}/${this.processId}~process@1.0/now`,
       {
         method: 'HEAD',
-        signal: AbortSignal.timeout(5000), // 5 second timeout
+        signal: AbortSignal.timeout(this.hbTimeoutMs),
       },
     )
-      .then((res) => {
+      .then(async (res) => {
         if (res.ok) {
+          if (minSlot !== undefined) {
+            const slotRes = await this.compute({
+              path: 'at-slot',
+              json: false,
+            });
+            const slot = Number(slotRes);
+            if (slot < minSlot) {
+              return false;
+            }
+          }
           this.isHyperBeamCompatible = true;
           return true;
         }
@@ -213,6 +235,7 @@ export class HB implements Hyperbeam {
           throw new Error(`Failed to fetch path: ${res.statusText}`);
         }
         const body = await res.text();
+        console.log('body', body);
         return body as T;
       }
     } catch (error) {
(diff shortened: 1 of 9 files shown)


arielmelendez:
@coderabbitai review

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,177 @@
...
+    this.checkHyperBeamPromise = this.checkHyperBeamCompatibility();
+  }
+
+  async meta(): Promise<JSONValue> {
+    const url = new URL(`${this.url}/${this.processId}~process@1.0/meta`);
+    url.searchParams.set('require-codec', 'application/json');
+    url.searchParams.set('accept-bundle', 'true');
+    const res = await fetch(url);
+    return res.json() as Promise<JSONValue>;
+  }
```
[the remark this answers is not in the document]
atticusofsparta: [f299228](https://github.com/ar-io/ar-io-sdk/pull/563/commits/f29922814d9c13619350f94b3a28bbb1c14a9e86)

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,177 @@
...
+    );
+    if (json) {
+      url.searchParams.set('require-codec', 'application/json');
+      url.searchParams.set('accept-bundle', 'true');
+    }
+    const res = await fetch(url);
+    const jsonResult = await res.json();
+    const body = (jsonResult as { body: T }).body;
+    return body as T;
+  }
```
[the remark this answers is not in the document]
atticusofsparta: [f299228](https://github.com/ar-io/ar-io-sdk/pull/563/commits/f29922814d9c13619350f94b3a28bbb1c14a9e86)

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,177 @@
...
```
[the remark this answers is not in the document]
atticusofsparta: [f299228](https://github.com/ar-io/ar-io-sdk/pull/563/commits/f29922814d9c13619350f94b3a28bbb1c14a9e86)

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,225 @@
...
+    if (this.isHyperBeamCompatible !== undefined) {
+      return Promise.resolve(this.isHyperBeamCompatible);
+    }
+
+    const result = fetch(
+      // use /now to force a refresh of the cache state, then compute when calling it for keys
+      `${this.url.toString()}/${this.processId}~process@1.0/now`,
+      {
+        method: 'HEAD',
+        signal: AbortSignal.timeout(5000), // 5 second timeout
```
atticusofsparta: make configurable in constructor

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,225 @@
...
+      });
+
+    this.checkHyperBeamPromise = result;
+
+    return result;
+  }
+
+  async fetchHyperbeamPath<T extends JSONValue>({
+    path,
+    json = false,
```
atticusofsparta: default to true

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,251 @@
...
+  }): Promise<T>;
+  compute<T extends JSONValue>({
+    path,
+    json,
+  }: {
+    path: string;
+    json?: boolean;
+  }): Promise<T>;
+  checkHyperBeamCompatibility(params?: { minSlot?: number }): Promise<boolean>;
+}
```
dtfiedler: nit: we don't need an interface (we won't have multiple implmeentatinos) - so just have a single class with these methods and you can call them from ARIO.

general takeaway: we "over interfaced" early on with ARIO and ANT classes when we didn't need to

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,232 @@
...
+    // refetch if min slot is provided
+    if (minSlot !== undefined) {
+      this.isHyperBeamCompatible = undefined;
+      this.checkHyperBeamPromise = undefined;
+    }
+    if (this.checkHyperBeamPromise !== undefined) {
+      return this.checkHyperBeamPromise;
+    }
+
+    if (this.isHyperBeamCompatible !== undefined) {
```
arielmelendez: It's not clear to me how this case would ever be hit since you're not removing the reference to the promise before it resolves. You effectively end up with the same logical outcome since you always resolve the same value as you would have cached here. My initial suggestion to you was to add a function `isHyperBeamAvailable(): Promise<boolean>` that simply returns the cached promise and to have `checkHyperBeamCompatibility` ALWAYS reassign the old promise reference.

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,232 @@
...
+        this.isHyperBeamCompatible = false;
+        return false;
+      });
+
+    this.checkHyperBeamPromise = result;
+
+    return result;
+  }
+
+  async fetchHyperbeamPath<T extends JSONValue>({
```
arielmelendez: Nit: this function is a little busy because it combines the fetching and the result parsing rather than using composition to deal with the result differently depending on caller needs, after fetching, on the outside of the function.

arielmelendez (approved):
Some minor comments worth considering but seems to get us what we need for the moment.

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,232 @@
...
+        }
+        return jsonResult.body as T;
+      } else {
+        this.logger.debug('Fetching path as text', { path });
+        const res = await fetch(url);
+        if (!res.ok) {
+          throw new Error(`Failed to fetch path: ${res.statusText}`);
+        }
+        const body = await res.text();
+        return body as T;
```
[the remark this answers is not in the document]
atticusofsparta: Not certain it is but the idea is to match the generic with is required to be returned and text is always a string so if the generic is number, bool, null, or object then it wouldn't match

src/common/hyperbeam/hb.ts
atticusofsparta: I was attempting to take an axios style approach here where it automatically parses the json. I wanted the fetching to be abstracted and the "device methods" like compute and process to just worry about the url structuring.

I think this was partially in response to some coderabbit comments regarding duplication of code - but I guess I could make this a fetch wrapper sort of and just return the whole result.

This is easily changeable in the future.

src/common/hyperbeam/hb.ts
```diff
@@ -0,0 +1,232 @@
...
+      path: url.toString(),
+    });
+  }
+
+  /**
+   * calls the process device /now function, which evaluates the current process state pulling new messages
+   * to get the latest state
+   *
+   * @param path - The path to the hb state
+   * @param json - Whether to return the result as JSON, defaults to true
```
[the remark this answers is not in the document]
atticusofsparta: I guess the implication is that either way its returning as the generic. In practice thats not exactly true though. IDK what a good way to handle this would be. Perhaps a selector function passed in as an argument that parses the raw body/response?

PR #565, opened 2025-10-16
Add error handling and logging for AO epoch distribution message parsing

kunstmusik:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for epoch data processing. Malformed messages are now handled gracefully instead of causing application failures, with the system logging errors and continuing to process remaining data without interruption.

### DIFF
--- a/src/utils/arweave.ts
+++ b/src/utils/arweave.ts
@@ -189,6 +189,9 @@ export const getEpochDataFromGqlFallback = async ({
       continue;
     }
     for (const message of messageResult?.Messages ?? []) {
+      if (!message.Data) {
+        continue;
+      }
       try {
         const tags: { name: string; value: string }[] = message.Tags;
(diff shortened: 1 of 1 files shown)


arielmelendez:
@coderabbitai review

src/utils/arweave.ts
```diff
@@ -188,20 +189,31 @@ export const getEpochDataFromGqlFallback = async ({
       continue;
     }
     for (const message of messageResult?.Messages ?? []) {
-      const data = JSON.parse(message.Data);
```
arielmelendez: The real fix that we need here is that we have to filter out messages that don't contain Data (e.g. patch messages).

atticusofsparta (dismissed):

atticusofsparta (approved):

PR #568, opened 2025-10-21
fix(spawn): update spawn ant utility to use configured ao client for ant registry process

atticusofsparta:


src/utils/ao.ts
```diff
@@ -259,7 +259,11 @@ export async function spawnANT({
   // check the ACL for the owner
   const antRegistry = ANTRegistry.init({
     signer,
-    processId: antRegistryId,
+    process: new AOProcess({
```
kunstmusik: should this use hyperbeamURL too?

arielmelendez (approved):

PR #573, opened 2025-11-07
Composite ARIO provider for legacy <> hyperbeam

atticusofsparta:


src/common/io-providers.ts
```diff
@@ -0,0 +1,1246 @@
...
+ */
+export interface ARIOReadProvider {
+  getBalance(params: { address: WalletAddress }): Promise<number>;
+  getPrimaryNameRequest(params: {
+    initiator: WalletAddress;
+  }): Promise<AoPrimaryNameRequest>;
+  getPrimaryName(
+    params: { address: WalletAddress } | { name: string },
+  ): Promise<AoPrimaryName>;
+}
```
dtfiedler: suggestion: use the existing AoARIORead interface and implement that. for any functions not yet supported throw an Error

src/common/io.ts
```diff
@@ -226,6 +227,7 @@ export class ARIOReadable implements AoARIORead, ArNSNameResolver {
   protected paymentProvider: TurboArNSPaymentProviderUnauthenticated; // TODO: this could be an array/map of payment providers
   protected logger = Logger.default;
   protected hb: HB | undefined;
+  protected readProvider: ARIOCompositeReadProvider;
```
dtfiedler: suggestion - keep this class as is, it's essentially the traditional CU based ARIORead implementation. suggest creating a separate ARIOCompositeRead implmenetation that accepts an array of read providers (`AoARIORead` and `HyperbeamARIORead` implementations) and implements the interface by calling the providers sequentially for each API. 

example composite provider in ar-io-node for reference: https://github.com/ar-io/ar-io-node/blob/develop/src/resolution/composite-arns-resolver.ts

dtfiedler: this composite could also accept some kv cache to store data client side too, that's one thing we never successfully added to the existing ARIO class

PR #600, opened 2026-03-24
fix(cli): normalize argv[1] path separators for Windows compatibility

JonnieSparkles:
## Problem

On Windows, the CLI silently exits without output when run via `node lib/esm/cli/cli.js` or the globally installed `ar.io` command. No error is thrown, nothing is printed — including `--help`.

The root cause is the conditional at the bottom of `cli.ts` that determines whether to call `program.parse()`:

```typescript
if (
  process.argv[1].includes('bin/ar.io') ||
  process.argv[1].includes('cli/cli')
) {
  program.parse(process.argv);
}
```

On Windows, `process.argv[1]` uses backslashes, e.g.:
```
C:\Users\jp\AppData\Roaming\npm\node_modules\@ar.io\sdk\lib\esm\cli\cli.js
```

Neither `'bin/ar.io'` nor `'cli/cli'` match because the path contains `cli\cli`, so `program.parse()` is never called and the CLI exits silently.

## Fix

Normalize the path separators before the check:

```typescript
const normalizedArgv1 = process.argv[1].replace(/\\/g, '/');

if (
  normalizedArgv1.includes('bin/ar.io') ||
  normalizedArgv1.includes('cli/cli')
) {
  program.parse(process.argv);
}
```

## Testing

Confirmed on Windows (PowerShell) that:
- Without the fix: `node lib/esm/cli/cli.js --help` produces no output
- With the fix: `--help` and commands like `get-arns-record --name ardrive --mainnet` work correctly

No impact on macOS/Linux — backslashes don't appear in those paths so the replace is a no-op.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross‑platform CLI startup by normalizing path separators when detecting how the tool was invoked, making initialization and command parsing more reliable across Windows, macOS, and Linux.


atticusofsparta:
resolved in https://github.com/ar-io/ar-io-sdk/pull/681

PR #628, opened 2026-05-12
feat(solana): return live delegation balances from read methods

vilenarios:
## Summary

Makes `getGatewayDelegates`, `getDelegations`, and `getAllDelegates` return the **live** delegate balance (last-settled principal + accumulator-derived pending rewards) instead of the raw `Delegation.amount` field, which is stale-by-design between settlements.

## Why this matters

`distribute_epoch` advances the per-share accumulator `Gateway.cumulative_reward_per_token` and does **not** update every `Delegation.amount` (it would be O(delegates) per epoch — would blow the compute budget). `Delegation.amount` is only refreshed when the delegate next interacts (`increase`/`decrease`/`redelegate`/`claim_from_leaving`) or when someone calls the permissionless `compound_delegation_rewards`.

The network portal (and every downstream consumer of these SDK methods) was reading the raw field — so a delegate who stakes once and lets it sit sees their displayed balance **never grow between epochs**, even though they're earning rewards. It only jumps when something triggers settlement.

This is corrected on the SDK read side without any on-chain change (the on-chain accumulator pattern is canonical and correct — see `INVARIANTS.md` in the contracts repo).

## Changes

- **`constants.ts`**: add `REWARD_PRECISION = 1e18` (mirrors the on-chain constant in `programs/ario-gar/src/state/mod.rs`).
- **`deserialize.ts`**: stop skipping `reward_debt` in `deserializeDelegation` and `cumulative_reward_per_token` in `deserializeGateway`; surface both as `bigint` on the internal return types. The public `AoGateway` type is unchanged.
- **`delegation-math.ts`** (new): `computeLiveDelegationBalance` helper that ports the on-chain `settle_delegate_rewards` math to TypeScript bigint, including the overflow-safe quotient/remainder split and the `u64::MAX` saturation cap.
- **`io-readable.ts`**:
  - New private `getGatewayAccumulators(operatorAddresses[])` helper — batch-fetches gateways via `fetchEncodedAccounts`, returns `Map<address, bigint>` of per-gateway accumulators.
  - `getGatewayDelegates`: fetches the single referenced gateway, applies live math to each delegation's `delegatedStake`.
  - `getDelegations` and `getAllDelegates`: collect all referenced gateways, batch-fetch them once, apply live math.
- **`delegation-math.test.ts`** (new): 7 unit tests — zero principal, no-delta short-circuit, monotonic-accumulator guard, single-epoch reward (10% rate), non-zero `rewardDebt` snapshot, large multiplier precision, and `u64` saturation.

## API compatibility

Return shapes unchanged. Callers see strictly better numbers (live ≥ stored). No new methods exposed.

## Cost

One additional `getMultipleAccounts` per call when delegations span multiple gateways. Negligible for the per-gateway reader (single RPC); modest constant factor for the global reader.

## Test plan

- [x] `yarn build:esm` — clean
- [x] `yarn lint:check` — clean (biome formatting applied)
- [x] `yarn test:unit src/solana/delegation-math.test.ts` — 7 subtests pass

## Summary by CodeRabbit

* **New Features**
  * Delegation balances now automatically account for earned rewards in real-time, providing accurate, up-to-date balance information.
  * Added reward accumulator tracking to support live balance calculations based on current gateway reward data.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ar-io/ar-io-sdk/pull/628)

### DIFF
--- a/src/solana/constants.ts
+++ b/src/solana/constants.ts
@@ -30,6 +30,12 @@ export const MPL_CORE_PROGRAM_ID: Address = address(
 export const TOKEN_DECIMALS = 6;
 export const ONE_TOKEN = 1_000_000; // 1 ARIO = 1,000,000 mARIO
 export const RATE_SCALE = 1_000_000;
+/**
+ * Scaling factor for the per-share reward accumulator
+ * `Gateway.cumulative_reward_per_token` in `ario-gar`. Must match the
+ * `REWARD_PRECISION` constant in `programs/ario-gar/src/state/mod.rs`.
+ */
+export const REWARD_PRECISION = 1_000_000_000_000_000_000n; // 1e18
 
 // =========================================
 // PDA Seeds — ario-core
--- a/src/solana/delegation-math.test.ts
+++ b/src/solana/delegation-math.test.ts
@@ -1,0 +1,91 @@
+import { strict as assert } from 'node:assert';
+import { describe, it } from 'node:test';
+
+import { REWARD_PRECISION } from './constants.js';
+import { computeLiveDelegationBalance } from './delegation-math.js';
+
+describe('computeLiveDelegationBalance', () => {
+  it('returns the raw amount when there is no accumulator delta', () => {
+    const live = computeLiveDelegationBalance({
+      delegatedStake: 10_000_000_000, // 10k ARIO
+      rewardDebt: 1234n,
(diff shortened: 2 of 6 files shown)


vilenarios:
@coderabbitai review

PR #632, opened 2026-05-15
fix: convert Solana delegate reward share from basis points to percent

fedellen:
The Solana gateway read path was returning delegateRewardShareRatio as raw basis points from on-chain storage, so gateways configured for 90% could appear as 9000% in the SDK/UI. The change converts the deserialized value back to percent so the SDK read path matches the write path and the public API.


vilenarios (approved):
Lookin good!

PR #73, opened 2024-05-02
fix(utils): manally convert from b64 to b64url to avoid web polyfill i…

dtfiedler:
…ssues

Somewhat related to polyfill issues not handling `base64url` correctly - https://github.com/feross/buffer/pull/314/files

PR #639, opened 2026-05-27
fix(solana): don't sign ANT writes against an untrusted auto-detected program

vilenarios:
## Summary

**Security finding (Codex), client-side.** `ANT.init` auto-detects the ANT program id from the asset's `ANT Program` Attributes-plugin trait when `antProgramId` is omitted (BYO-ANT, ADR-016 / BD-100). That trait is read from Metaplex Core asset / RPC data and is **not authenticated**.

For a **writeable** client the detected value became the instruction `programAddress` across every write method (`setRecord`, `setController`, `transfer`, …) while the caller was included as a **signer**. So a malicious asset owner — or a spoofed/compromised RPC response — could route a victim's signed transaction to an attacker-controlled Solana program and inherit signer/writable privileges over the included accounts. The CLI/SDK write path hits this by default whenever `antProgramId` is omitted.

## Fix — harden the signing path, keep BYO-ANT

BYO-ANT stays supported, but only via **explicit opt-in** for non-canonical programs:

- **`resolveWriteAntProgram` (`mpl-core.ts`)** — the trust gate:
  - an explicit `antProgramId` (caller-supplied) is always honored;
  - an auto-detected value is trusted only when it's absent (`null`) or equals the canonical `ARIO_ANT_PROGRAM_ID`;
  - a detected **non-canonical** value with no explicit opt-in **throws** instead of silently signing against it.
- **`ANT.init`** — the **read** path still auto-detects freely (it never signs); the **write** path runs the detected value through `resolveWriteAntProgram`.
- **`io-writeable.ts` primary-name path** — only honor a detected program when it equals canonical or this client's explicitly-configured `this.antProgram`, else fall back to the configured program. (This path only derives a READONLY validation account — the instruction targets the canonical core program — so it's lower risk, but gated for defense in depth.)

## Tests

`src/solana/mpl-core.test.ts` (8 new):
- pure `resolveWriteAntProgram` cases (canonical / null / non-canonical-throws / explicit-opt-in / explicit-wins);
- `ANT.init` end-to-end against a spoofed `ANT Program` trait: **rejected for writes**, **allowed for reads**, **bypassed by explicit opt-in**.

All four CI gates pass locally: `lint:check`, `format:check`, `build` (tsc), `test:unit` (236 pass).

## Drive-by: pre-existing tsc break fixed

`EscrowAntState` / `EscrowTokenState` typed `version` as `{ major, minor, patch }`, but the generated decoder (and the on-chain account) expose a `u8` schema-version `number`. This broke `yarn build` on `solana` already (the `solana` branch is excluded from push-triggered CI, so it went unnoticed). Corrected to `number`. Unrelated to the security fix but required for a green build.

## Contracts-side follow-up (separate repo)

The on-chain `ant_program == ario_ant::ID` pin (ario-core primary-name path, from the on-chain half of the same Codex report) currently blocks heterogeneous BYO-ANT end-to-end. Whether to relax it is a separate `ar-io-solana-contracts` decision (flagged with the team) and is **not** part of this PR.

### DIFF
--- a/src/common/ant.ts
+++ b/src/common/ant.ts
@@ -33,6 +33,13 @@ import type { ANTRead, ANTWrite } from '../types/ant.js';
  * (BYO-ANT) and falls back to the canonical default. Callers who already know
  * the program (e.g. immediately after `ANT.spawn`) can pass it explicitly to
  * skip the lookup.
+ *
+ * SECURITY: the `ANT Program` trait is untrusted asset/RPC data. For a
+ * read-only client it's used as-is, but a writeable client (signer present)
+ * will NOT sign against an auto-detected *non-canonical* program — it throws
+ * unless you opted in by passing `antProgramId` explicitly. This prevents a
+ * malicious asset or spoofed RPC response from redirecting your signed
+ * transaction to an attacker-controlled program.
  */
 export type ANTConfig = {
   processId: string;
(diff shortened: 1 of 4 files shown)


vilenarios:
Rebased onto `solana` (was 4 merges behind: #640, #641, #644, #645).

Conflict surface was small and benign:
- `src/solana/escrow.ts` — the drive-by `version` typing fix in this PR's body is now a **redundant no-op**: #641 shipped the same correction (`EscrowAnt['version']` / `EscrowToken['version']`) on 2026-05-27. Rebase kept the (identical) lines from one side; final state on `solana` is unchanged.
- `src/solana/io-writeable.ts` — disjoint methods: this PR's gating is in `_buildPrimaryNameValidationAccounts`; `solana`'s `50c92b29` change is in `reassignName`/`releaseName` (ant_asset wiring for contracts PR #73). Auto-rebased cleanly.

Local gates green on the rebased tip:
- `yarn lint:check` — clean
- `yarn format:check` — clean
- `yarn build` — tsc clean
- `yarn test:unit` — 0 failing

The security fix itself is unchanged from the original commit. Ready for review/merge once green CI confirms.

PR #643, opened 2026-05-29
feat(solana): single-tx primary-name replace + ArnsRecord auto-migrate; pin solana-contracts 0.4.0

atticusofsparta:
## Summary

Pins `@ar.io/solana-contracts` to exactly `0.4.0` and adds three pieces of defensive SDK logic so callers get the new behavior without operational surprises:

### 1. `_buildMigrateArnsRecordIxIfNeeded` — auto-migrate stale records

Every ARNS write path (`extendLease`, `increaseUndernameLimit`, `setPrimaryName`, `requestPrimaryName`, `requestAuctionedName`, `reassignName`, `releaseName`, `claimReleasedName`, `upgradeRecord`, `syncAttributes`) now pre-checks the on-chain `ArnsRecord` and prepends `migrate_arns_record` if its stored `name_hash` (bytes 8..40) doesn't match the expected value. Lets the SDK keep working transparently against records imported under earlier schemas.

### 2. `_buildRemoveExistingPrimaryNameIxs` — single-tx primary-name change

`setPrimaryName` / `requestPrimaryName` now prepend a `remove_primary_name` ix when the signer already has one, so the on-chain `request_and_set_primary_name` doesn't reject with `MustRemoveExistingPrimaryName` (0x1786 / 6022). Single-tx atomic replace.

**Legacy fallback (updated in commit 174ccef):** when the signer's existing PrimaryName has no paired `PrimaryNameReverse` PDA (pre-#159 import state), the original PR silently skipped the remove and hoped `request_and_set` would handle it. It doesn't — the 6022 check is unconditional. The fallback now throws a typed Error pointing at `yarn workspace @ar-io/migration-import backfill:primary-name-reverse` so operators get an actionable message instead of an opaque on-chain failure.

### 3. ACL self-heal on `ANT.transfer`

When transferring an ANT whose owner ACL entry was never seeded (e.g. pre-ACL-era marketplace acquisitions), the SDK now bootstraps the missing ACL config + page and records the entry in the same tx so the on-chain `transfer` handler can successfully remove it. Was: throw `AclEntryNotFound`.

## Tradeoffs

**RPC fan-out:** every ARNS write path now does at least one extra `fetchEncodedAccount` for migrate detection. `setPrimaryName` does up to two extras (forward PrimaryName check + reverse PDA check on the existing primary). For high-throughput callers this is real RPC budget. Post-mainnet migration, when every record is at the current schema, these checks are pure waste. Worth a follow-up env flag (e.g. `ARIO_SDK_SKIP_MIGRATE_CHECK=1`) to short-circuit after the cluster is fully migrated. Not blocking — single-digit-ms latency hit per write.

## Tests

- **Unit:** 237/237 pass.
- **Localnet** (skipped without RPC env): 3 new cases in `io-writeable.localnet.test.ts`:
  - atomic primary-name replace: covers the headline new behavior end-to-end, asserts old reverse cleaned + new reverse initialized in one tx.
  - legacy primary-name state throws actionable error: forges a forward-without-reverse via `import_account` (migration authority), asserts client-side rejection matches the new error message + on-chain code 6022, no tx submitted.
  - `ANT.transfer` regression guard for the normal path. Heal-path proper isn't covered (needs an unseeded-ACL fixture that's hard to construct on a fresh localnet) — TODO comment in the test.
- **Codecov:** below the 80% target. The added paths are now covered by the localnet tests above; standalone unit coverage of `_buildMigrateArnsRecordIxIfNeeded` would require RPC mocking on a private method and was judged not worth the maintenance burden (covered transitively by every test that touches an ArnsRecord).

## Reviewer checklist

- [ ] Run `yarn sdk-e2e:up && yarn test:localnet:io-write` against a working localnet to confirm the three new cases pass (my local bootstrap is blocked on a pre-existing authority-mismatch when deploying over devnet-cloned programs — environmental, not PR-related).
- [ ] Sanity-check the migrate-record offset assumption: nameHash lives at bytes 8..40 in the current `ArnsRecord` schema. Verified against `@ar.io/solana-contracts/lib/arns/accounts/arnsRecord.js` codegen.
- [ ] Confirm `migration_active` is still ON on the cluster you'll deploy this against — `_buildMigrateArnsRecordIxIfNeeded` returns a `migrate_arns_record` ix that requires it. Phase-7'd clusters will see opaque failures here; consider documenting that constraint or adding a similar fail-fast.
- [ ] CodeRabbit auto-review skipped because the PR targets `solana` not `main` (same pattern as #642/#161). Reply `@coderabbitai review` on the PR to trigger.

## Summary by CodeRabbit

* **Bug Fixes**
  * ANT transfer now automatically heals missing permission entries for smoother transactions.
  * Primary name operations now properly remove existing entries before changes, ensuring consistency.
  * Added automatic cleanup for legacy primary name inconsistencies with helpful error guidance.
  * Records are automatically migrated when needed during standard operations.

* **Tests**
  * Extended regression test coverage for primary name and transfer scenarios.

* **Chores**
  * Pinned dependency version for stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ar-io/ar-io-sdk/pull/643?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


vilenarios:
@coderabbitai review

vilenarios:
Worked through the three nitpicks. Verified each against current code; one was already correct, one is out of scope, one is now tracked.

### ① tx CU/size when bundling heal + prep + transfer + cleanup ixs

**Valid concern, skipped for this PR.**

Worst-case bundle is 2 (oldOwnerHealIxs) + 2 (newOwnerDest.prepIxs) + 1 (transferIx) + N (cleanupIxs, one per ex-controller). For typical ANTs (N ≤ 5) this fits comfortably in PACKET_DATA_SIZE (1232 B); for heavily-controlled ANTs it could exceed.

Not fixing in this PR because:
- The cleanup-ix bundling is pre-existing — this PR only adds at most 2 heal ixs to the front, an incremental capacity hit, not a new class of failure.
- Splitting into multiple txs (heal-prep → transfer → cleanup) trades atomicity for capacity: a successful heal-prep followed by a failed transfer leaves an orphan ACL entry that the heal-on-next-attempt path doesn't currently handle gracefully.
- The atomic single-tx contract is the better default for the common case. The edge case (many controllers + unseeded ACL + transfer) is rare.

If this becomes a real problem in production, the right fix is a `simulateTransaction`-gated split with explicit rollback semantics — a separate refactor, not a nitpick fix.

### ② page-PDA / pageIdx binding in the heal branch

**Invalid — the binding is structurally enforced by construction.**

`resolveDestinationAclAccounts` (registry:200-205) returns:
```ts
const [aclPagePda] = await getAclPagePDA(params.user, pageIdx, this.antProgram);
return { aclConfigPda, aclPagePda, pageIdx, prepIxs };
```

`buildRecordIx` (registry:304-308) derives:
```ts
const [aclPage] = await getAclPagePDA(params.user, params.pageIdx, this.antProgram);
```

Both call the same `getAclPagePDA(user, pageIdx, antProgram)` with the same inputs. The heal branch passes `dest.pageIdx` to `buildRecordIx`, so `oldOwnerAclPagePda` and the `aclPage` inside the record ix are derived from identical `(user, pageIdx, programAddress)` tuples — divergence would require modifying `getAclPagePDA` to be non-deterministic.

Adding an assertion would be a tautology; adding a test would exercise `getAclPagePDA`'s determinism rather than the heal logic. Skipped.

### ③ heal-path test missing

**Valid — tracked as #646.**

Opened https://github.com/ar-io/ar-io-sdk/issues/646 with acceptance criteria + two implementation strategies (fixture helper or state surgery). The TODO comment in the test now references the issue (commit 259c813).

PR #680, opened 2026-06-08
fix(solana): source demand-factor settings from program constants

atticusofsparta:
## What

Source `getDemandFactorSettings` from the on-chain program constants (`@ar.io/solana-contracts/arns`) instead of hand-copied literals.

## Why

The literals had drifted from the program:
- **Down-adjustment was `25` (2.5%)** while the program uses `DEMAND_FACTOR_DOWN_ADJUSTMENT = 985_000` → 0.985× → **1.5% (`15`)**. The ar.io pricing docs already say 1.5%; the SDK was the outlier.
- **`maxPeriodsAtMinDemandFactor` returned the live consecutive-period counter** (`df.consecutivePeriodsWithMinDemandFactor`) instead of the `MAX_PERIODS_AT_MIN_DEMAND_FACTOR` threshold.

## How

Import the constants (generated from the Rust source-of-truth) and derive the rate fields, so they can no longer drift:

```ts
demandFactorUpAdjustmentRate:   Number(((DEMAND_FACTOR_UP_ADJUSTMENT   - scale) * 1000n) / scale), // 50
demandFactorDownAdjustmentRate: Number(((scale - DEMAND_FACTOR_DOWN_ADJUSTMENT) * 1000n) / scale), // 15
maxPeriodsAtMinDemandFactor:    MAX_PERIODS_AT_MIN_DEMAND_FACTOR,                                   // 7
```

`upRate` stays `50` (unchanged); `downRate` corrects `25 → 15`.

## Test

- New `demand-factor-settings.test.ts` pins the constant values/types and the derivation — **3/3 pass**.
- Full offline solana unit suite: **290/290 pass**.
- `tsc --noEmit`: clean.

## Dependency

Consumes `@ar.io/solana-contracts@1.0.1` (published from ar-io/ar-io-solana-contracts#110, which exports the arns constants). The release is published and pinned in `package.json` + `yarn.lock` — **no longer blocked**; tests and typecheck above run against the published package.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #682, opened 2026-06-09
fix(solana): price returned-name premium against the cluster clock

atticusofsparta:
## Problem

`getTokenCost` / `getCostDetails` quoted the **wrong price for returned names** (names in the post-expiry Dutch auction).

The auction applies a premium multiplier that decays linearly **50x → 1x** over 14 days. The SDK derived the elapsed time from `Date.now()`:

```ts
const now = Date.now();
const elapsed = now - returned.startTimestamp;
const multiplier = ((50n * dur - 49n * el) * scale) / dur;
cost = (cost * multiplier) / scale;
```

But the on-chain handler measures elapsed time against the **Solana cluster clock** (`Clock::get().unix_timestamp`) at the landing slot — not the client wall clock. The cluster clock lags real wall-clock by seconds-to-minutes, so the SDK over-estimated elapsed time, under-estimated the remaining premium, and **quoted low versus what the chain charges**. At ~3.5x/day of decay, even a small clock gap moves the price.

## Fix

- Add a read-only `getClusterUnixTimestampSeconds()` (`getSlot` → `getBlockTime`) that reads the same stake-weighted clock the on-chain `Clock` sysvar exposes — **no signer, no transaction**.
- Use it (converted to ms) instead of `Date.now()` for the returned-name premium, so the quote matches on-chain.
- Fetch it **once** per `getTokenCost` and also use it to detect a **stale demand factor** (a demand period has rolled on-chain but no crank has landed to roll the stored factor). When detected, emit a `logger.debug` warning and point callers at the simulated writeable `getCostDetails` for a byte-exact quote. We deliberately don't replicate the roll client-side (the authoritative multi-period/fee-halving math lives in the Rust program and isn't faithfully reproducible read-only).
- The coarse 180-day gateway-discount tenure gate keeps `Date.now()` — sub-minute drift is immaterial at day granularity, so it avoids an extra RPC round trip.

## Test

- Full offline solana unit suite: **368/368 pass**.
- `tsc --noEmit`: clean.

## Notes

This is the targeted correctness fix for returned-name pricing. It's independent of #680 (which fixes the *reported* `getDemandFactorSettings`, a separate introspection method that isn't in the pricing path). Both can land independently.

### DIFF
--- a/src/solana/io-readable.ts
+++ b/src/solana/io-readable.ts
@@ -368,6 +368,35 @@ export class SolanaARIOReadable {
   }
 
   /**
+   * The Solana cluster's `unix_timestamp` (in SECONDS), read read-only.
+   *
+   * On-chain handlers measure time against the cluster clock
+   * (`Clock::get()?.unix_timestamp`), NOT the client wall clock. For any
+   * pricing math that must agree with what the program will charge (e.g. the
+   * Dutch-auction premium for returned names, whose elapsed time gates the
+   * multiplier), the SDK must compare against this same clock — `Date.now()`
+   * drifts from `unix_timestamp` (the cluster clock famously lags real
+   * wall-clock by seconds-to-minutes), which would make the SDK over-estimate
+   * elapsed time and therefore UNDER-quote the premium.
+   *
+   * Sourced via the latest slot's block time (`getSlot` → `getBlockTime`),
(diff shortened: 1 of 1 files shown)


atticusofsparta:
Superseded by #683, which combines this with the other ArNS pricing-accuracy fix on a single branch off the latest `alpha`. Closing in favor of #683.

PR #685, opened 2026-06-16
fix(solana): filter finalize_gone candidates by leave-window eligibility

vilenarios:
## Problem

The Solana epoch cranker spams `finalize_gone` simulation failures. On AR.IO mainnet epoch go-live we saw **~280 `LeaveWindowNotExpired` (Anchor error 6079) reverts per 10 minutes** from a single operator's cranker.

Root cause: `getLeavingGateways()` / `getGoneGateways()` (`io-readable.ts`) return **every** `status === Leaving` gateway with no eligibility filter — by design, per their own comment ("over-returning not-yet-eligible Leaving gateways is safe — the tx just no-ops/reverts"). A cranker loop that calls `finalizeGone()` per result (e.g. ar-io-observer's `epoch-cranker.ts` phase-3 GC) therefore burns a tx attempt **and** logs a full simulation-failure stack trace on every gateway whose leave window hasn't elapsed, every tick. It costs no SOL (sim/revert) and doesn't wedge anything, but it's continuous noise that buries real cranker/observer logs.

The decoder already carries the eligibility data and discards it: a raw `Gateway` has `leaveTimestamp` (leave-window end, on-chain seconds) and `totalDelegatedStake` — the exact two preconditions the on-chain `finalize_gone` enforces (`programs/ario-gar/src/instructions/gateway.rs::finalize_gone` reverts before the window elapses, and won't GC a gateway with live delegations).

## Fix

Add `getFinalizableGoneGateways(now)` that returns only `Leaving` gateways where:
- the leave window has elapsed: `now >= leaveTimestamp`, and
- no delegated stake remains: `totalDelegatedStake === 0`.

This mirrors the existing expiry-filtered discovery pattern (`getExpiredVaults(now)`, `getExpiredArnsRecords(now)`). The unfiltered `getLeavingGateways()` is unchanged for callers that want raw discovery; cranker loops should prefer the filtered method. A follow-up will switch ar-io-observer's phase-3 GC to it.

## Tests

`prune-discovery.test.ts` — 6 new cases (inject encoded `Gateway` accounts, stubbed RPC):
- ✅ Leaving + window elapsed + no delegations → included
- ✅ Leaving + window not elapsed → excluded (the 6079 case)
- ✅ boundary `now === leaveTimestamp` → eligible
- ✅ Leaving + delegations remain → excluded
- ✅ Leaving + no `leaveTimestamp` → excluded
- ✅ Joined → excluded

Full solana unit suite passes; biome clean.

### DIFF
--- a/src/solana/io-readable.ts
+++ b/src/solana/io-readable.ts
@@ -3021,6 +3021,52 @@ export class SolanaARIOReadable {
   }
 
   /**
+   * Enumerate `Leaving` Gateway PDAs that are ACTUALLY eligible for
+   * `finalizeGone` at `now` — i.e. their leave window has fully elapsed
+   * (`now >= leaveTimestamp`) AND no delegated stake remains
+   * (`totalDelegatedStake === 0`). These are exactly the two preconditions the
+   * on-chain `finalize_gone` enforces: it reverts `LeaveWindowNotExpired`
+   * before the window elapses, and will not GC a gateway that still has live
+   * delegations (programs/ario-gar/src/instructions/gateway.rs::finalize_gone).
+   *
+   * {@link getLeavingGateways} over-returns *every* `Leaving` gateway and
+   * relies on those reverts happening on-chain. A cranker that calls
+   * `finalizeGone` per result therefore burns a tx attempt — and logs a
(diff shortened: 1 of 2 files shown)


vilenarios:
@coderabbitai review

PR #687, opened 2026-06-17
fix(solana): finalize_gone eligibility must use the full leave window

vilenarios:
## Corrects #685

`getFinalizableGoneGateways(now)` (added in #685) filtered on `leaveTimestamp <= now` only. But the on-chain `finalize_gone` GC window (`programs/ario-gar/src/instructions/gateway.rs`) is:

```
eligibleAt = leaveTimestamp + GATEWAY_LEAVE_PERIOD (90 days)
                            + 7 × max(leaveEpochDuration_snapshot, epoch_settings.epoch_duration)
```

So #685 returned gateways still inside their ~97-day cooldown. A cranker calling `finalizeGone` on them hits the exact revert the method was supposed to prevent (`LeaveWindowNotExpired` / 6079).

**Verified live on mainnet** (running the #685 build in an isolated observer on an operator box): `getFinalizableGoneGateways` returned **33 "eligible"** → **0 finalize_gone successes, 100% reverted**. The spam continued unchanged.

## Fix
Compute the real window:
- `GATEWAY_LEAVE_PERIOD_SECONDS = 90 * 86_400` (mirrors `ario-gar` `GATEWAY_LEAVE_PERIOD`).
- `+ 7 * max(leaveEpochDuration, current epoch_duration)` — read current `epoch_duration` via `getEpochSettings()`; on read failure, fall back to the per-gateway snapshot (the 90-day term still dominates).
- Keep the `totalDelegatedStake === 0` precondition.

## Tests
`prune-discovery.test.ts` rewritten for the real window — incl. a **regression test** for "past 90d but not past 90d + 7 epochs → NOT eligible" (the case the old filter wrongly admitted), plus boundary, recently-left, delegated-stake, no-timestamp, and Joined cases. Full solana unit suite: **321 pass / 0 fail**; biome clean.

## Rollout
Release → bump `@ar.io/sdk` in ar-io-observer (the consumer already calls `getFinalizableGoneGateways(now)` via #100, no code change) → rebuild observer image → deploy. Expectation on the current young mainnet: **0 finalizable gateways → 0 finalize_gone attempts → no spam.**

## Release Notes

* **Bug Fixes**
  * Improved gateway finalization eligibility calculations by properly accounting for epoch duration and aligning system behavior with on-chain requirements, ensuring gateways transition correctly through their lifecycle at the appropriate time.

### DIFF
--- a/src/solana/io-readable.ts
+++ b/src/solana/io-readable.ts
@@ -324,6 +324,21 @@ function paginate<T>(
  * const record = await ario.getArNSRecord({ name: 'ardrive' });
  * ```
  */
+
+/**
+ * Fixed leave cooldown (seconds) a `Leaving` gateway must wait before
+ * `finalize_gone` can GC it. Mirrors the on-chain `GATEWAY_LEAVE_PERIOD`
+ * (90 days) in ar-io-solana-contracts (`programs/ario-gar/src/lib.rs`).
+ */
+const GATEWAY_LEAVE_PERIOD_SECONDS = 90 * 86_400; // 7_776_000
+
+/**
+ * Epoch-retention multiplier in the on-chain finalize_gone window
+ * (`leave_timestamp + GATEWAY_LEAVE_PERIOD + 7 * epoch_duration`), mirroring
+ * the GAR-006 epoch retention.
+ */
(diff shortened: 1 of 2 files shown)


vilenarios:
@coderabbitai review

PR #693, opened 2026-06-29
fix(arns): resolve txId from ANT record in resolveArNSName

kempsterrrr:
## Summary

`SolanaARIOReadable.resolveArNSName` returned a hardcoded `txId: ''` (and a hardcoded `ttlSeconds: 3600`) for **every** name. It looked up the ArNS record — which only yields the ANT `processId` — and stopped there, leaving the `// TODO: resolve undername via ANT program` hop unimplemented. Callers therefore never received the transaction id a name actually resolves to.

This wires in the missing second read:

- Parse the **undername** from the requested name: the apex `@` for a bare name (`ardrive`), otherwise everything before the final `_`-delimited base name (`logo_ardrive` → undername `logo`, base `ardrive`).
- Load the ANT via `SolanaANTReadable.fromAsset(...)` so the correct per-ANT program id is read from the asset's `ANT Program` attribute (ADR-016 / BD-100) rather than assuming the canonical program.
- Fetch the undername record and map its `transactionId` → `txId`.
- Return the record's real `ttlSeconds` and `priority` instead of placeholders, and echo back the full requested `name`.
- Throw when the undername has no record on the ANT, instead of silently returning an empty `txId`.

The ANT fetch is split into a protected `resolveAntRecord` seam so the resolver's parsing/mapping logic is covered by fast unit tests without RPC (the fetch itself is already covered by `SolanaANTReadable`'s tests).

### Before
```jsonc
// resolveArNSName({ name: 'ardrive' })
{ "name": "ardrive", "txId": "", "type": "permabuy", "processId": "US517…", "ttlSeconds": 3600, "undernameLimit": 10 }
```

### After
```jsonc
{ "name": "ardrive", "txId": "<ANT record target>", "type": "permabuy", "processId": "US517…", "ttlSeconds": <record ttl>, "priority": <record priority>, "undernameLimit": 10 }
```

## Testing

- New unit tests in `src/solana/resolve-arns-name.test.ts` cover: apex txId mapping, TTL/priority pass-through, undername resolution + full-name echo, multi-segment undername parsing, and the missing-record throw.
- `yarn test:unit` — **407 passing, 0 failing**.
- `yarn tsc -p tsconfig.json --noEmit` — clean.
- `biome check` / `biome format` — clean.

## Checklist

- [x] My PR is against the correct branch (`alpha`)
- [x] My commit messages follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) style
- [x] My commit messages clearly explain the reasons for the changes
- [x] I have followed the architecture guidelines
- [x] I have run lint and format checks
- [x] I have run all tests
- [x] I have checked for existing issues/PRs to avoid duplicates


src/solana/io-readable.ts
```diff
@@ -3310,19 +3313,59 @@ export class SolanaARIOReadable {
...
+   * Fetch the undername record from the ANT backing `processId`.
+   *
+   * Uses {@link SolanaANTReadable.fromAsset} — the canonical resolution-path
+   * factory — so the correct ANT program id is read from the asset's
+   * `ANT Program` Attributes-plugin entry (ADR-016 / BD-100) rather than
+   * assuming the canonical program. Returns `undefined` when the ANT has no
+   * record for the undername. Split out so the resolution mapping in
+   * {@link resolveArNSName} stays unit-testable without an RPC.
+   */
+  protected async resolveAntRecord(
```
atticusofsparta: we don't need and extra function for this, can just inline it in resolveArNSRecord.

kempsterrrr: Done — removed `resolveAntRecord` and inlined the `fromAsset(...).getRecord(...)` calls into `resolveArNSName` (f440d7b).

I'd only extracted it as a test seam, which isn't how this repo tests reads anyway — so I reworked the unit test to stub at the RPC layer instead (serializing a real `ArnsRecord` + `AntRecord` and driving the actual two-hop fetch), mirroring `primary-name.test.ts`. That also covers the fetch path the seam was hiding, which should clear the Codecov patch-coverage flag.

Two related changes I made while inlining, for visibility:

- The parser now reuses the existing `splitPrimaryName` instead of a bespoke split. The old hand-rolled split didn't lowercase, so a mixed-case name (`Logo_ArDrive`) derived the wrong PDAs and silently failed to resolve — `splitPrimaryName` lowercases like the contract. It also unifies the split rule to first-`_` / `splitn(2, '_')` (was last-`_`); identical for all valid names since neither base names nor undernames may contain `_`.
- To avoid a circular import (readable ← writeable), `splitPrimaryName` moved from `io-writeable.ts` to a shared `name.ts`. No behavior change; it isn't part of the public API.

Happy to split the parser change into its own PR if you'd rather keep this one to just the txId fix.

src/utils/arns.ts
```diff
@@ -0,0 +1,43 @@
...
+  const ix = lower.indexOf('_');
+  if (ix === -1) {
+    return { isUndername: false, baseName: lower, undername: null };
+  }
+  return {
+    isUndername: true,
+    baseName: lower.slice(ix + 1),
+    undername: lower.slice(0, ix),
+  };
+}
```
atticusofsparta, line 43: We already have a utils file i think, this belongs there. In fact we may already have a util that does this.

kempsterrrr: Moved it to `src/utils/arns.ts` with a co-located test and barrel export (c0f3577); deleted `src/solana/name.ts` and repointed the importers. The `splitPrimaryName` unit tests moved out of `primary-name.test.ts` into `arns.test.ts`. No behavior change.

On "we may already have a util for this" — I checked and there isn't a shared one, but the same split is duplicated inline as a `baseNameOf` closure in `io-readable.ts` (~L2531 and ~L2622, both `n.toLowerCase().split('_')`).

Two optional follow-ups I left out to keep this change minimal — happy to fold either into this PR or do a separate one, your call:

1. **Consolidate the duplication** — replace both `baseNameOf` closures with `splitPrimaryName`. Behavior-identical for valid names (they only diverge on invalid 3+-segment input the contract rejects).
2. **Rename `splitPrimaryName` → `splitArNSName`** — moving it into the utils barrel makes it public API (`utils/index.js` is re-exported from the package entry), and the name reads oddly for a general ArNS splitter now that it also drives resolution. Purely cosmetic; touches ~8 call sites.

atticusofsparta (approved):

PR #697, opened 2026-07-20
docs: rename ArNS heading to "ar.io Name System"

vilenarios:
ArNS is the **ar.io** Name System. This heading was the last `Arweave Name System` reference anywhere in the published docs — every other occurrence (18 of them, including the glossary) already says "Ar.io Name System".

```diff
-### Arweave Name System (ArNS)
+### ar.io Name System (ArNS)
```

## ⚠️ This changes a published URL

This README is the source for the ar.io SDK reference on docs.ar.io (via `generate-sdk-docs`), and the page **filename is derived from this heading**. So merging this moves:

```
/sdks/ar-io-sdk/arweave-name-system-arns   →   /sdks/ar-io-sdk/ar-io-name-system-arns
```

To avoid breaking that link, the docs side needs a follow-up **after** this merges and `generate-sdk-docs` runs:

1. Add a `redirects.mjs` entry from the old path to the new one.
2. Update the link in `content/build/access/arns.mdx:53`.

I'll open that docs PR once this lands — flagging it here so the sequencing is explicit and the URL doesn't 404 in between.

Companion docs work: ar-io/docs#124.

https://claude.ai/code/session_01NUArqiyP9GbfJrqNFVkUBr

## Summary by CodeRabbit

* **Documentation**
  * Updated the ArNS documentation heading to refer to the ar.io Name System (ArNS).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1296,7 +1296,7 @@ const delegates = await ario.getAllDelegates({
 
 </details>
 
-### Arweave Name System (ArNS)
+### ar.io Name System (ArNS)
 
 #### `resolveArNSName({ name })`
 


vilenarios:
@coderabbitai review

vilenarios:
@coderabbitai Good catch on the slug change — agreed on the substance, but the fix can't ship in this repo, and there's a wrinkle worth recording.

**1. The files you reference live in a different repo.** `redirects.mjs` and `content/build/access/arns.mdx` are in [ar-io/docs](https://github.com/ar-io/docs), not `ar-io-sdk`. This README is only the *source* — `generate-sdk-docs` fetches it and derives the page (and its filename) from the heading.

**2. The link update must come after this merges, not with it.** `generate-sdk-docs` has to run against the new heading before `/sdks/ar-io-sdk/ar-io-name-system-arns` exists. Repointing `arns.mdx` first would break the link checker on a page that isn't generated yet.

**3. `redirects.mjs` won't actually save the old URL.** The docs site builds with `output: "export"`, and Next.js drops redirects in static export — the production build logs it explicitly:

```
⚠ rewrites, redirects, and headers are not applied when exporting your application, detected (redirects).
```

So a `redirects.mjs` entry would be inert in production. The realistic options are a gateway/CDN-level redirect, or accepting that the old path falls through to the site's 404 → `/learn` fallback.

Sequencing is tracked in ar-io/docs#124. Flagging option 3 for the maintainers to decide, since it changes whether this heading fix is worth the URL churn.

PR #707, opened 2026-08-14
fix(solana): apply sortBy in paginate instead of only echoing sortOrder

vilenarios:
## The bug

The Solana `paginate` helper sliced its input and returned `sortOrder` in the result **without ever sorting**. `sortBy` was not accepted at all.

```ts
function paginate<T>(items, params?: { cursor?; limit?; sortOrder? }) {
  const page = items.slice(startIdx, startIdx + limit);   // slice only
  return { items: page, ..., sortOrder: params?.sortOrder ?? 'asc' };  // echoed
}
```

So this call:

```ts
ario.getGateways({ limit: 25, sortBy: 'operatorStake', sortOrder: 'desc' })
```

returned gateways in **registry insertion order**, sliced to the limit, while the response asserted `sortOrder: 'desc'`.

Dropping the key is bad. **Reporting a sort that did not happen is worse**, because it defeats the check a careful caller would make before trusting the ranking. In the report that surfaced this, several low-stake sibling gateways dominated a diagnostic sample that was believed to be stake-ranked.

`PaginationParams<T>` declares `sortBy?: SortBy<T>`, so the **type advertised a capability the implementation did not provide.** This is a contract violation rather than a documentation gap.

Scope: `paginate` has **18 call sites** in the Solana read client, so this affected every paginated read there, not only `getGateways`.

## The fix

| Behaviour | Why |
|---|---|
| `sortBy` applied **before** slicing | so `limit` returns the true top N, not the first N in registry order |
| Dot paths resolve (`settings.fqdn`) | `SortBy<T>` is `NestedKeys<T>`, so nested keys are already permitted by the type |
| `bigint` compared as `bigint` | stake and balance fields are mARIO and routinely exceed `Number.MAX_SAFE_INTEGER`; coercing collapses distinct values into ties and mis-ranks exactly the leaderboard this sort serves |
| Nullish sorts **last in both directions** | a missing field is absent, not smallest — burying it beats floating it to the top of a `desc` list |
| Input array copied, never sorted in place | callers may hold their own reference |
| **No `sortBy` → ordering unchanged** | existing callers are unaffected |

## Tests

Nine cases in `src/solana/paginate.test.ts`, including the two that matter most:

- **sort before slice** — `{ limit: 2, sortBy: 'stake', sortOrder: 'desc' }` returns the genuine top two
- **bigint precision** — two stakes that provably collapse to the same value under `Number()` (`9007199254740992n` and `9007199254740993n`) still rank correctly. The test asserts that precondition explicitly, so it fails loudly if someone reintroduces a `Number()` coercion.

```
paginate: 9/9 pass
full unit suite: 465 tests, 0 fail
tsc --noEmit: clean
```

`paginate` is exported for the test only; it is **not** re-exported from any package index, so the public surface is unchanged.

## Deliberately not included

`PaginationParams` also declares **`filters`**, which this helper ignores in exactly the same way. Same class of bug. I kept it out so this fix stays reviewable — happy to follow up with it if you'd like.

## Summary by CodeRabbit

* **New Features**
  * Added optional sorting to paginated results, including ascending and descending order.
  * Supports sorting by nested properties and values such as numbers, text, booleans, and large integers.
  * Null or missing values are consistently placed last.
  * Sorting occurs before result limits and works with cursor-based pagination.

* **Bug Fixes**
  * Input collections remain unchanged when sorting is applied.


vilenarios:
Addressed in `88f6818`. Verified against the type definitions first, and the finding was correct on both counts.

**`PaginationResult<T>` already declares `sortBy?: SortBy<T>`** — I checked, and my previous commit never populated it. So a caller inspecting `result.sortBy` saw `undefined` even after requesting a sort. That is the same defect as the original bug pointing the other way: metadata that does not describe what happened. Now echoes the key actually applied.

**`SortBy<T>` type-checks cleanly at all 18 call sites.** I had used `sortBy?: string` to sidestep generic friction; that caution was unnecessary, so the parameter now matches `PaginationParams<T>` and callers get key checking rather than an unconstrained string. `valueAtPath` still takes a plain string, since `SortBy<T>` is a union of dot-path literals.

Tests assert both directions — a requested sort reports its key, an unsorted call claims nothing.

`paginate`: 10/10 · full suite: 466 tests, 0 fail · `tsc --noEmit` clean.

vilenarios:
`filters` is now implemented too — `f196cd7`.

It was declared on `PaginationParams` and discarded exactly like `sortBy`: a filtered query returned the **full unfiltered set**, with `totalItems` and `hasMore` describing data the caller had explicitly excluded.

- a scalar matches by equality; an array matches if **any** entry does, so values are alternatives within a field while separate fields must all match
- **bigint fields compare by value against a plain number** — callers write numbers, mARIO amounts are bigint, and matching on type alone would silently return nothing
- filtering runs **before** sorting and slicing, so `totalItems`, `hasMore` and the page all describe the filtered set

Also fixed an import-ordering failure I introduced: `biome`'s `lint:check` enforces alphabetical order and I had inserted `SortBy` next to `PaginationResult`. `tsc` and the tests both passed locally, so I had not run `lint:check` — my gap, not CI's. I now run the full CI set locally.

**All 8 CI jobs green** on `f196cd7` (build, lint:check, format:check on 18.x and 20.x, plus unit tests).

`paginate`: 14/14 · full suite: 470 tests, 0 fail

PR #712, opened 2026-08-23
perf(solana): cut redundant RPC calls in the read path

vilenarios:
## What

Four RPC-efficiency fixes in the Solana read path, found by tracing every RPC call site and measuring against devnet. Each is a separate commit.

| Commit | Fix |
|---|---|
| `f05d72f6` | Coalesce concurrent reads behind in-flight caches (+ cluster-clock memo) |
| `031f101c` | Batch gateway metadata in discovery; drop a dead account read |
| `f8959664` | Fetch account chunks in a bounded pool instead of serially |
| `5ebf33c9` | Test portability: drop `mock.timers` (needs Node 20.4; we support 18) |

### 1. Caches stored values, not in-flight promises

The short-TTL caches in `SolanaARIOReadable` only ever helped *sequential* callers. A concurrent burst all miss before any of them fills the cache — and the burst is exactly what the caches exist for (a price table quoting every row at once).

Measured on devnet, ten concurrent `getGasEstimate()`:

| | RPC calls | bytes | fee queries |
|---|---|---|---|
| sequential | 14 | 21 KiB | 3 |
| concurrent (before) | 43 | 200 KiB | 30 |

The burst also tripped an HTTP 429. Nearly all of it was `getRecentPrioritizationFees` at ~6.6 KiB a call.

Two correctness details in the new helper: rejections are evicted rather than cached, so one transient 429 can't poison a key for its whole TTL; and eviction is identity-guarded, so a late-settling stale request can't delete a newer entry for the same key. A `keep()` predicate preserves `getCachedAccount`'s "misses are not cached" rule while still letting a burst share the single lookup that finds the miss.

Also memoizes `getClusterUnixTimestampSeconds` (two *sequential* round trips, called on every `getTokenCost`). TTL is deliberately 1s so a per-second Dutch-auction UI stays accurate; staleness is directionally safe (an older clock over-quotes the premium rather than under-quoting).

### 2. Dead and serialized reads in funding discovery

`fetchGatewayMeta` ran one `getAccountInfo` per delegation, serially — 120 delegations meant 120 sequential round trips. Only the account's *existence* is consulted, but that bit selects between two `minDelegationAmount` values and so changes the drawdown sort, so it can't just be deleted; it is now one batched `getMultipleAccounts` per 100.

`fetchOperatorStake` fetched an account and then returned `null` on every code path. Removing it changes no output — operator stake never appeared in the discovered set, so `fundAsOperator` has always required explicit `sources`. Real operator-stake discovery is a behavior change and is left for its own PR.

### 3. Chunks fetched one at a time

Four bulk readers batched at 100 accounts but awaited each chunk before starting the next. Devnet's 648 gateways: ~210ms serialized vs ~110ms pooled; mainnet's registry widens the gap.

Capped at 4 concurrent, not unbounded — an unbounded `Promise.all` turns one bulk read into a burst as wide as the registry, which is how you earn a 429. `getGateways` also gained the `withRetry` its sibling paths already had, since raising concurrency without it would turn a transient 429 into a hard failure.

## Why this can't regress consumers

Every touched field is `private`, or `protected` with an unchanged signature. Nothing new is exported from `src/solana/index.ts`.

- **Public type surface unchanged.** Built `lib/types` before and after and diffed. Every delta is a private member or a doc comment. The one new file (`concurrency.d.ts`) is neither re-exported from the barrel nor reachable through the `exports` map.
- **Behavioral equivalence, deterministically.** Recorded a devnet RPC transcript, then replayed the same 27 read scenarios against it before and after. **0 output diffs, 27.5% fewer RPC calls.** Replay makes the comparison immune to chain state and fee-market drift. Includes the multi-chunk `getGateways({limit:250})` and 150-mint ANT reads where ordering is load-bearing.
- **CLI unchanged.** Built the pre-change commit in a separate worktree and ran 11 read commands against devnet on both. Output byte-identical.

## Tests

+13 tests (485 → 498, 0 failures). The one skip is pre-existing.

The `getANTSummaries`/`getANTStates` readers index results positionally (`accounts[i * 3]`), so flattening in completion order would silently attribute one ANT's config to another mint. The ordering test drives chunks that finish in **reverse** order, and was mutation-checked: reversing the flattening makes it fail.

`discoverFundingSources` had no coverage at all, so its tests were written against the *original* implementation first — the shape test passed before the refactor and still does; the batching test failed against the original with 122 `getAccountInfo` calls for 120 delegations and now sees 1.

## Notes for the reviewer

- **Pre-existing, not addressed here:** `getGasEstimate` and `getGarGasEstimate` share `_rentCache` but compute `rentKey` differently (`sum(bytes)` vs `sum + realloc + count`), so the two schemes can collide. Behavior is unchanged by this PR; worth a follow-up.
- A separate PR will follow for the `rpc-circuit-breaker.ts` bugs reported by the network-portal team (fallback traffic bypassing the throttle; an advertised rps-limit disabling backoff). Both reproduced locally.

## Summary by CodeRabbit

- **Performance**
  - Improved bulk Solana account reads with bounded concurrent fetching while preserving result order.
  - Reduced duplicate RPC requests by sharing in-flight reads and briefly caching cluster-time data.
  - Batched gateway metadata requests during funding-source discovery.

- **Behavior Changes**
  - Operator stake is no longer discovered automatically; it remains available through explicit sources.
  - Preserved retry handling and missing-account behavior.

- **Bug Fixes**
  - Prevented incorrectly matched account data when requests complete out of order.

### DIFF
--- a/src/solana/ant-readable.test.ts
+++ b/src/solana/ant-readable.test.ts
@@ -138,3 +138,139 @@ describe('SolanaANTReadable request batching', () => {
     assert.equal(counts.gai, 0, 'no single-account reads');
   });
 });
+
+// ---------------------------------------------------------------------------
+// Chunk ordering under bounded concurrency
+//
+// getANTSummaries and getANTStates read each mint's accounts POSITIONALLY out
+// of the flattened batch result (`accounts[i * 3]`, `accounts[i * 2]`). Once
+// the 100-account chunks stopped being awaited one-at-a-time, chunk completion
+// order stopped matching request order — so if the flattening ever followed
+// completion order instead of input order, one ANT's config would silently be
+// attributed to another mint. These tests make the chunks finish out of order
+// on purpose and assert every mint still gets its own data back.
+// ---------------------------------------------------------------------------
+
+const antMint = (n: number) => bs58.encode(Buffer.alloc(32, n));
+
+/**
+ * Stub whose chunk responses resolve in REVERSE order (the later the chunk,
+ * the sooner it settles), and which echoes the requested PDA back so each
+ * decoded account can be traced to the mint it was derived from.
+ */
+function outOfOrderChunkRpc(
+  counts: { gma: number },
+  configFor: (pda: string) => unknown | null,
+) {
+  let chunkIndex = 0;
+  return {
+    getProgramAccounts: () => ({ send: async () => [] }),
+    getAccountInfo: () => ({ send: async () => ({ value: null }) }),
+    getMultipleAccounts: (addrs: unknown[]) => {
+      const myIndex = chunkIndex++;
+      return {
+        send: async () => {
+          counts.gma++;
+          // Later chunks return first.
+          await new Promise((r) =>
+            setTimeout(r, Math.max(0, 30 - myIndex * 10)),
+          );
+          return {
+            value: (addrs as string[]).map((a) => configFor(a)),
+          };
+        },
+      };
+    },
+  };
+}
+
+describe('ANT bulk reads — chunk ordering', () => {
+  it('keeps each mint matched to its own accounts across several chunks', async () => {
+    // 60 mints x 3 PDAs = 180 accounts = 2 chunks of <=100.
+    const mints = Array.from({ length: 60 }, (_, i) => antMint(i + 1));
+    const counts = { gma: 0 };
+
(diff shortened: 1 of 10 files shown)


vilenarios:
@coderabbitai review

PR #713, opened 2026-08-23
fix(solana): make the RPC throttle back off when it actually matters

vilenarios:
Fixes reported by the ar.io network-portal team, which has already shipped a local workaround (it no longer uses `createCircuitBreakerRpc`). I reproduced every claim before touching anything; two of the six turned out to be worth fixing here, and the rest are triaged at the bottom.

## 1. Fallback traffic bypassed the throttle entirely

opossum only emits `failure` on the **closed**-circuit path. Once the circuit is OPEN it emits `reject` and calls the fallback directly (`lib/circuit.js`, the `!this.closed && !this.pendingClose` branch), so nothing the fallback returned — 429s included — ever reached the backoff handler. The rate gate stayed pinned at its ceiling for exactly as long as the primary was down, which is when backpressure matters most.

Measured before the fix: a fallback serving nothing but 429s absorbed **58 requests in 2.5s** at a ceiling of 20 r/s. The fallback is now wrapped so its own failures feed the same backoff.

Both transports still share one rate gate, so backing off on the fallback also slows primary retries. That's the conservative direction; a per-transport gate would be a reasonable follow-up.

## 2. A provider-advertised rps limit disabled the backoff

`min(ceiling, advertised * SAFETY)` resolves to the ceiling itself whenever the provider advertises a limit above ours. `api.mainnet-beta.solana.com` sends `x-ratelimit-rps-limit: 250` against a default ceiling of 10, so the header path was a **no-op** there.

A 429 always means slow down, so an advertised limit may now only pull the rate *down*: `max(MIN_RATE, min(advertised * SAFETY, current * AIMD))`.

**Trade-off:** a 429 can no longer jump the rate *up* toward a provider's advertised capability, so recovery to that rate goes through the slow additive path. That's the intent — a rate-limit response should never raise throughput — but it is a behavior change.

## 3. `fallbackUrl` is now optional

Requiring it was a forcing function: a consumer with one paid endpoint had nowhere to point it but a public RPC, so a degraded primary quietly became traffic aimed at shared infrastructure. With no fallback configured, none is registered — the circuit still protects the primary and open-circuit calls fail fast rather than being routed elsewhere.

Backward compatible: widening required → optional keeps every existing caller compiling. The `.d.ts` diff is exactly one line.

## Testing notes — this seam is easy to test wrongly

Worth reading before reviewing the tests, because two of my own attempts were wrong first:

- **The existing 429 test sets `volumeThreshold: 100` specifically to keep the circuit CLOSED.** That's why it never caught bug 1. The new test drives the OPEN path.
- **Bug 2 cannot be caught by an absolute request count.** The per-429 cooldown dominates throughput and masks the rate — measured, broken and fixed builds were *indistinguishable* that way, and my first test for it passed against the broken code. The test uses `retry-after: 0` to remove the cooldown, leaving throughput governed purely by the adaptive rate, and asserts against the known-good headerless case rather than a fixed number so it doesn't depend on machine speed. With the cooldown removed the bug is stark: **68 requests versus 9**.

Both new throttle tests fail against the previous implementation with clear diagnostics, and pass after. The six pre-existing tests are unchanged and still pass.

## Also in here

`getInfo()`'s cost is now documented: it spends two account reads while returning `Ticker`, `Name`, `Denomination`, `Handlers` and `LastCreatedEpochIndex` as fixed literals — only `totalSupply`, `protocolBalance` and `epochSettings` come from chain. Documented rather than given a separate static-metadata accessor; permanently growing the public surface to save two reads in a marginal case is the worse trade.

## Reported but deliberately not fixed

- **10s opossum timeout.** The stated premise doesn't hold: mainnet has **2650** ARIO token accounts and a full `getBalances` scan moves ~**0.67 MB**, landing a few seconds in — comfortably inside 10s. (The reported "2349 accounts / 1.13 MB" is the *devnet* figure.) The real concern next door is trip sensitivity — `volumeThreshold: 3` with `errorThresholdPercentage: 25` means one timeout in four requests opens the circuit for 60s. Worth revisiting, but not by inflating the timeout on a false premise, and not as a default change smuggled into a bugfix PR.
- **`getAllDelegates` always fetching accumulators.** A `{ liveBalances?: boolean }` opt-out is cheap and additive, but only pays off if a consumer genuinely wants addresses without balances. Happy to add it if the portal still wants it.

## Verification

- Full suite: 484 tests, 0 failures. `lint:check`, `format:check`, `build` all clean.
- Public type surface: one intended line (`fallbackUrl?`). `io-readable.d.ts` otherwise unchanged.
- Blast radius: one source file plus a docstring.

## Summary by CodeRabbit

* **New Features**
  * Circuit breaker configurations can now operate without a fallback endpoint.
  * Requests fail fast with clearer behavior when no fallback is available and the circuit is open.
  * Rate limiting adapts more reliably to HTTP 429 responses and provider-advertised limits.

* **Documentation**
  * Added documentation describing on-chain fields, static metadata, and RPC costs.

### DIFF
--- a/src/solana/io-readable.ts
+++ b/src/solana/io-readable.ts
@@ -796,6 +796,22 @@ export class SolanaARIOReadable {
   // Protocol info
   // =========================================
 
+  /**
+   * Protocol info: on-chain supply/balance figures plus static protocol
+   * metadata.
+   *
+   * COSTS TWO ACCOUNT READS (ArioConfig + EpochSettings), which is worth
+   * knowing because only some of the result comes from chain:
+   *
+   * - From chain: `totalSupply`, `protocolBalance`, `epochSettings`.
+   * - Fixed literals: `Ticker` (`'ARIO'`), `Name` (`'AR.IO'`),
+   *   `Denomination` (`6`), `Handlers` (`[]`), `LastCreatedEpochIndex` (`0`).
+   *
+   * If all you need are the literals — a ticker or denomination for display —
+   * don't call this; hard-code them or read {@link TOKEN_DECIMALS}. Two RPC
+   * round trips for a string constant is a poor trade, and callers have
+   * reached for this method not realising that is what they were paying.
+   */
   async getInfo() {
     const [[configPda], [epochSettingsPda]] = await Promise.all([
       getArioConfigPDA(this.coreProgram),
--- a/src/solana/rpc-circuit-breaker.test.ts
+++ b/src/solana/rpc-circuit-breaker.test.ts
@@ -254,6 +254,130 @@ describe('createCircuitBreakerRpc', () => {
   });
 });
 
+describe('createCircuitBreakerRpc — adaptive throttle', () => {
+  const servers: Array<{ close: () => Promise<void> }> = [];
+
+  afterEach(async () => {
+    await Promise.all(servers.map((s) => s.close()));
+    servers.length = 0;
+  });
+
+  /** A port with nothing listening, so the primary transport always fails. */
+  async function deadUrl(): Promise<string> {
+    const s = await createStatusMockServer(() => ({ statusCode: 200 }));
+    await s.close();
+    return s.url;
+  }
+
+  /** Drive `concurrency` callers for `ms`, swallowing the expected errors. */
+  async function drive(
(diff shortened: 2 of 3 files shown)


vilenarios:
@coderabbitai review

PR #714, opened 2026-08-23
fix(solana): don't report zero funding sources when discovery fails

vilenarios:
Small follow-up, found while differential-testing #712's discovery changes against staging v2.

**Stacked on #712** — it touches the same functions, so it targets that branch rather than `alpha`. Merge #712 first and this rebases cleanly.

## The bug

Both `getProgramAccounts`-backed discovery scans wrapped their entire body in `try { … } catch { return [] }`. A transient rate-limit was therefore indistinguishable from *"this wallet has no delegations"*. The planner would build a plan from incomplete sources and quietly under-fund it — no error, no log, on the path that pays for ArNS purchases.

Not hypothetical. Differential-testing discovery against staging v2, a burst of 429s from public devnet turned a wallet holding **seven delegations into one reporting none**. It took a second run with wider spacing to establish the stake had been there the whole time. That is exactly what a consumer would hit under load, except they would have no second run to tell them.

## The fix

The empty-list fallback exists for a real case — many public RPCs disable `getProgramAccounts` outright — so it is kept for **permanent** errors, and now logs a warning rather than vanishing.

**Transient** failures (429/5xx/network) are retried through `withRetry` — which these scans were missing entirely, unlike every equivalent read path in `io-readable`/`ant-readable` — and then surfaced. A caller can retry or pass explicit `sources`. Being told *zero* when the truth is *unknown* is the one outcome that is never acceptable here.

The batched gateway-metadata reads get the same `withRetry` treatment.

## Behavior change

A purchase that previously proceeded balance-only after silently losing its delegation sources will now throw instead. That is the intent — under-funding from phantom-empty discovery is worse than a loud, retryable failure — but it is a change, and it is the one thing worth arguing about in review.

## Verification

- 500 tests, 0 failures; `lint:check`, `format:check`, `build` clean.
- Public type surface identical to the parent branch.
- Both new tests **mutation-checked**: reverting the handler to swallow everything makes the transient-failure test fail. One asserts the transient path retries then throws; the other asserts an unsupported-`getProgramAccounts` RPC still degrades gracefully to balance-only, without retrying a permanent error.

### DIFF
--- a/src/solana/funding-plan.test.ts
+++ b/src/solana/funding-plan.test.ts
@@ -722,4 +722,71 @@ describe('discoverFundingSources', () => {
     );
     assert.equal(counts.accountsRequested, 120);
   });
+
+  it('surfaces a transient RPC failure instead of reporting zero sources', async () => {
+    // The whole point: a 429 must not be indistinguishable from "this wallet
+    // has no delegations". Observed live against public devnet — a wallet
+    // holding seven delegations reported none, silently, and the planner would
+    // have under-funded from it.
+    const counts: DiscoveryCounts = {
+      getAccountInfo: 0,
+      getMultipleAccounts: 0,
+      getProgramAccounts: 0,
+      accountsRequested: 0,
+    };
+    const rpc = discoveryRpc(counts, { ataAmount: 1n, delegations: [] });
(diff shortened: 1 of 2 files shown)


vilenarios:
@coderabbitai review

PR #725, opened 2026-09-09
fix(solana): declare transaction-v1 support in ALT history scan

vilenarios:
## Why

Solana's **transaction v1** format (SIMD-0385, carrying the SIMD-0296 4096-byte ceiling) activates on mainnet at **epoch 1035, ~Sept 15 2026 01:20 UTC** (feature gate `txv1aq4pp281K9um3tnPgkfX8UqtFT6wcVW3hNezGLL`). Devnet and testnet are already live.

The upgrade is **additive** — legacy and v0 transactions are unchanged, and every transaction this SDK builds is v0 (`createTransactionMessage({ version: 0 })`). **The send path needs nothing.** The exposure is entirely on the read side, and we have exactly one.

`reclaimLookupTablesForSigner` discovers the ephemeral Address Lookup Tables it should clean up by replaying the signer's own transaction history. That `getTransaction` call declared `maxSupportedTransactionVersion: 0`, and a client declaring only v0 is answered with **JSON-RPC -32015** (`Transaction version (1) is not supported by the requesting client`) for every v1 transaction in range. Once a cranker signer has sent even one, the scan fails.

It failed badly, too. Unlike the reclaim loop directly below it, the discovery loop had **no error handling** — so a single unreadable signature propagated out of the function and aborted the whole pass, stranding the rent (~0.0126 SOL per table) of every ephemeral table that signer ever created.

## What changed

`src/solana/send.ts` — two independent guards, because either alone still loses the rent:

1. **Declare v1 support** (`maxSupportedTransactionVersion: 1`). Handles the common case. v1 carries no address table lookups — the format dropped them — so these entries contribute nothing to discovery; they simply must not throw.
2. **Make discovery best-effort**, matching the reclaim loop's existing contract. One signature we cannot read (pruned by the RPC, an unsupported future version, a transient failure) must not cost us the other 499.

Plus a doc correction on `MAX_TX_SIZE_BYTES`: 1232 is the **legacy/v0** cap. The 4096 ceiling applies only to v1, which this SDK cannot emit yet — `@solana/kit` still declares `SupportedTransactionVersion = Exclude<TransactionVersion, 1>`, so `createTransactionMessage` refuses it. v1 also drops ALTs entirely, so the compression in this file stays the only way past the inline account limit.

## Tests

`src/solana/send.test.ts` — three regression tests over a hand-written stub RPC, matching the repo convention (no mocking library):

- `maxSupportedTransactionVersion: 1` actually reaches `getTransaction`
- a `-32015` failure between two good signatures still discovers **both** tables
- an all-failing scan returns zeros instead of throwing

Each guard was reverted independently to confirm the tests bite:

| Reverted | Result |
| --- | --- |
| version bump only (try/catch kept) | 1 fail — `declares transaction-v1 support` |
| try/catch only (version kept) | 2 fail — both skip-and-continue tests |
| neither (this PR) | 6/6 pass |

## Verification

```
lint:check     0 errors (19 warnings — pre-existing, verified identical on a stashed tree)
format:check   clean
build:esm      clean
test:unit      544 tests, 543 pass, 0 fail, 1 skipped
```

The 1 skip is pre-existing (`escrow-token.test.ts:412`, an ADR-022 ABI test).

## Notes for the reviewer

- **Based on `alpha`, not `main`.** CONTRIBUTING.md says to branch from `main`, but that assumes the post-release state described in the same section ("After release, `main` should be merged back into `alpha`"). `alpha` is currently **70 commits ahead of `main`** with zero commits the other way, and `src/solana/send.test.ts` does not exist on `main` at all — it was added by alpha-only commit `70ac3031`. Branching from `main` would have produced an add/add conflict on that file. Worth a `main` ← `alpha` promotion at some point to restore the documented invariant.
- **Watch item, not addressed here:** the `'auto'` priority fee reads `getRecentPrioritizationFees`. If validators don't fold v1's `priorityFeeLamports` config field into that RPC's output, estimates could skew low on a v1-heavy network. Worth re-checking after epoch 1035.
- **Possible follow-up:** `prescribe_epoch` currently costs three confirmed transactions plus a finalized-visibility wait via `sendWithEphemeralLookupTable`. ~50 observer PDAs would fit v1's 4096 bytes, but land uncomfortably close to its 64-address inline cap — and v1 and ALTs are mutually exclusive. Needs a real account count before anyone commits to it, and is blocked on kit exposing v1 message creation regardless.

https://claude.ai/code/session_01FojVmRMP6hNyfjX1zkHHez

## Summary by CodeRabbit

- **Bug Fixes**
  - Lookup table reclamation now skips permanently unsupported transaction records while continuing history scans.
  - Transient RPC failures are retried according to the configured retry policy and reported when retries are exhausted, preventing incomplete results from appearing empty.
  - Retry behavior now consistently distinguishes permanent transaction-version incompatibilities from transient RPC failures.

- **Documentation**
  - Clarified the applicable transaction size limit for transactions created by the SDK.

### DIFF
--- a/src/solana/send.test.ts
+++ b/src/solana/send.test.ts
@@ -1,7 +1,7 @@
 import assert from 'node:assert/strict';
 import { describe, it } from 'node:test';
 import { address, generateKeyPairSigner } from '@solana/kit';
-import { sendAndConfirm } from './send.js';
+import { reclaimLookupTablesForSigner, sendAndConfirm } from './send.js';
 
 /**
  * A blockhash is only valid for ~150 blocks (~60s) from ISSUE, not from send.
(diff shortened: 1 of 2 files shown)


vilenarios:
@coderabbitai review

Re-requesting review: the previous run was rate limited after reviewing `131eab1b`, so commit `75ed0cb0` has not been looked at. That commit is the one worth a second opinion — it changes the error-handling policy of the ALT-discovery scan (retry transient failures, surface what survives, log-and-skip only permanent ones) to match the precedent set in ab1b7fc9.

vilenarios:
@coderabbitai review

vilenarios:
Fixed in `9dae3562` — good catch, and it's a real one.

`withRetry` classifies with `retryOptions.isRetryable` when a caller supplies one, but the post-exhaustion branch used the default `isRetryableError`. So a caller with a custom classifier got the two decisions made by *different* predicates: retried as transient, then classified as permanent, then skipped — silently truncating the candidate set. That's the same phantom-empty result this handler was added to prevent, reachable through the very `retryOptions` parameter it introduced.

Hoisted the predicate once (`const isTransient = retryOptions?.isRetryable ?? isRetryableError`) and used it for both decisions.

Added the test you asked for: a custom predicate marks the otherwise-permanent -32015 as retryable, and the scan must reject after exhaustion rather than skip. Mutation-checked — restoring the default classifier fails it.

Suite: 547 tests, 546 pass, 0 fail, 1 skipped (pre-existing).

vilenarios:
@coderabbitai review

vilenarios:
**Correction to a claim in this PR's description.**

The "Notes for the reviewer" section said `alpha` was ~70 commits ahead of `main` with zero commits the other way, and that `src/solana/send.test.ts` did not exist on `main`. Both were wrong.

I measured with a bare `main` ref in a long-lived local clone without fetching first. That ref was **72 commits stale**. Against `origin/main` (`75fa1de0`, v4.3.0, promoted in #724 on Aug 31), the real divergence at the time was small, and `send.test.ts` **did** exist there.

So the reasoning given for basing this branch on `alpha` rather than `main` was an artifact of a stale ref. CONTRIBUTING.md's "branch from `main`" is correct and the invariant it depends on is being maintained by the team. The outcome here was unaffected — the branch merged cleanly with green CI — but the justification in the description should not be taken as evidence about this repo's branch hygiene.

The correct check, for anyone reaching for it later:

```
git fetch origin && git rev-list --left-right --count origin/main...origin/alpha
```

PR #726, opened 2026-09-13
fix(solana): cover zeroed registry slots and assert cursor progress (D4/D4a)

vilenarios:
> **Depends on #728** — merge that `main` → `alpha` back-merge first. Until it lands, this PR shows `CONFLICTING` on release metadata only; a trial merge onto #728 is clean.

Fixes the client-side defect that silently stalled **mainnet epoch 542** (61,140 ARIO) and **staging epoch 817**, and that ADR-0032/0033 in the contracts repo explicitly do **not** address — this failure is entirely client-side.

## The bug — two of them, and the second is the nastier

`getRegistryGatewayPDAs` feeds both `tally_weights` and `distribute_epoch`:

```ts
const end = Math.min(startIndex + batchSize, count);  // (1) live registry.count
if (addr === zero) continue;                          // (2) emits NOTHING for zeroed slots
```

1. **Wrong bound.** Both on-chain loops walk to the epoch's `active_gateway_count`, which is **frozen at epoch creation**, while the registry is live. A `finalize_gone` mid-epoch drops `registry.count` below it and zeroes the trailing slots, so nothing is supplied for them.

2. **Zeroed slots were skipped, not filled.** Both loops iterate `remaining_accounts` and, for a slot whose address is the default pubkey, do `idx += 1; continue` — which **consumes one remaining account**. Omitting them starves the loop, so the cursor stops at the *first* zeroed slot. This is not a tail-only problem: any zeroed slot inside the range does it.

Either way the result is a **silent no-op** — the loop body never runs, the cursor is written back unchanged, and **the transaction succeeds**. There is no error to catch. That is why mainnet 542 and staging 817 both sat stalled while looking perfectly healthy in transaction history.

## The fix

**D4** — the range is capped by a new `slotCap` argument (both call sites pass `epoch.activeGatewayCount`), and zeroed slots yield a filler account. The filler is the program address itself: already present in the message as the invoked program, so it costs no account slot, and it is never dereferenced because the on-chain skip runs before any owner/writability check. It is attached **read-only** — see the review follow-up below for why that matters.

**D4a** — after each batched tally/distribute, re-read the epoch and throw if the cursor did not advance. Since the failure mode is a *successful* transaction, a stalled cursor is the only available signal; no try/catch or error-rate monitoring can see it.

Throwing is deliberate rather than returning a flag: it stops the tick **before** `createEpoch`. Creating the next epoch lets its tally overwrite the stuck epoch's weights and destroy its rewards permanently — that is exactly how mainnet epoch 540 lost its tail. A stalled epoch must freeze progression, not be advanced past.

The before-cursor is snapshotted into a local prior to sending rather than re-read from the epoch object, so an aliased or mutated object cannot silently defeat the comparison.

## Validated against live state

Running the real implementation against real registries, read-only:

| | slots needing coverage | **fixed** | **old behaviour** |
|---|---|---|---|
| mainnet 542 (cursor 633, active 647) | 14 | 14 accounts, 14 fillers | **0 accounts** |
| staging 818 (cursor 645, active 647) | 2 | 2 accounts, 1 filler | 1 account |

Mainnet 542 returning **zero** accounts is the silent no-op in one line. Staging 818 shows the subtler half — it supplies the real gateway but not the filler, stalling one slot short.

## Tests

4 new crank cases: a distribute that advances nothing, a tally that advances nothing, the normal advancing path, and that the batch range is capped by `activeGatewayCount`. Plus 6 `getRegistryGatewayPDAs` cases against a synthetic registry, and 2 that build and **compile** the real tally/distribute messages. **553 unit tests pass.**

The crank test mocks also needed fixing, and it is worth calling out: they **never mutated the epoch fixture**, so no call could ever advance a cursor — and `getEpochRaw` handed out the live object, letting callers alias harness state. Those mocks would have passed a crank that did nothing at all. They now advance cursors the way the chain does and return copies.

## Review follow-up (e5279083): the first version of this fix could not clear a D4 epoch

Final review found the filler was attached **writable**. The filler is the GAR program address, which is also the program the transaction invokes, and `@solana/kit` refuses to compile that message:

> This transaction includes an address (…) which is both invoked and marked writable. Program addresses may not be writable

So every batch crossing a zeroed slot, the exact case this PR targets, threw at compile time. It failed loudly rather than silently (the tick still stopped before `createEpoch`), but it could not have cleared mainnet epoch 543, which reaches a 13-slot zeroed tail tonight. The unit tests missed it because they only inspected the address list, and the crank mocks replace `tallyWeights` / `distributeEpoch`, so no real message was ever compiled.

- Fillers are now attached **read-only** through one helper, `batchRemainingAccounts`, used by both paths. Gateway PDAs stay writable.
- Read-only is safe on-chain: `tally_weights` and `distribute_epoch` both skip a zeroed slot *before* their owner and `is_writable` checks. It matches what `distribute-epoch-padded.mjs` used to clear mainnet 542 and staging 817/818.
- The new regression test builds the real tally and distribute instructions with fillers and compiles the message. Both cases fail on `cd7767a8` and pass on `e5279083`.
- Also fixed doc placement: the `slotCap` doc had been separated from `getRegistryGatewayPDAs` by `assertCursorAdvanced`, leaving the public method undocumented beneath a stale comment.

**Still not proven on-chain:** the SDK's own path has not yet sent a padded batch to a real cluster. The first natural opportunity is mainnet epoch 543 (window ends 2026-09-14T00:04Z): a read-only `simulateTransaction` of an SDK-built batch against its zeroed tail, before this reaches crankers.

## Checklist

- [x] My PR is against the correct branch — `alpha`, per CONTRIBUTING.md. **Merge #728 first:** it back-merges the 4.3.0 release into `alpha`; until then this PR shows `CONFLICTING`, on release metadata only (`CHANGELOG.md` / `package.json` / `src/version.ts`). A trial merge onto #728 is clean.
- [x] My commit messages follow conventional commits style
- [x] My commit messages clearly explain the reasons for the changes
- [x] I have followed the architecture guidelines
- [x] I have run lint and format checks (0 errors; remaining warnings are pre-existing unused-suppression notices)
- [x] I have run all tests — 553 unit tests, 0 failures
- [x] I have checked for existing issues/PRs to avoid duplicates — none; independent of #727, no linked issue

## Note for reviewers

`distribute-epoch-padded.mjs` in `solana-ar-io` was the manual stopgap used to clear mainnet 542 and staging 817. This PR is what stops it recurring — without it, every mid-epoch GC arms another silent stall that only a manual pre-flight catches.

https://claude.ai/code/session_01EsETyArrWiywkt763dUsVT

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved epoch processing reliability by detecting successful transactions that fail to advance tally or distribution progress.
  * Ensured batch processing respects the epoch’s active gateway count.
  * Preserved processing of empty or inactive gateway slots to prevent silent no-op failures.
  * Added coverage for cursor advancement, completion flags, and capped batch ranges.


vilenarios:
**On-chain evidence for the read-only filler fix (e5279083).**

Mainnet epoch 543 hit the exact D4 shape tonight: cursor stopped at **620/633**, live `registry.count` 620, slots 620–632 all zeroed. I built the `distribute_epoch` batch with this branch's code — `getRegistryGatewayPDAs(620, 30, activeGatewayCount)` → `distributeEpoch()` — and ran it through `simulateTransaction` against mainnet (read-only: `sendTransaction` stubbed to capture, `sigVerify: false`, nothing sent):

```
epoch 543: dist_idx=620/633 distributed=0 tallied=1 prescribed=1
batch from 620: 13 account(s), 13 filler(s)
filler metas read-only: yes
SIMULATION err=null units=28411
  Program log: Instruction: DistributeEpoch
  Program 89fNiiwgpFSPHKuqfNUkgYTYjtAJAhyqHjXmgXeppGpf consumed 28261 of 1399850 compute units
  Program 89fNiiwgpFSPHKuqfNUkgYTYjtAJAhyqHjXmgXeppGpf success
```

The same builder on `cd7767a8` cannot produce this transaction at all — it throws at compile time (*"Program addresses may not be writable"*). The deployed crankers, on the pre-#726 SDK, are sitting at 620 supplying zero accounts.

Earlier dry run on stuck epoch 540 (25 fillers) reached the handler and failed exactly where expected — `WeightsNotTallied (6048)` at a clobbered slot — confirming the message shape is accepted with duplicate read-only fillers.

HarukaMa:
Independently confirmed this failure on mainnet epoch **543**, using an observer running SDK **4.1.2**:

- Epoch snapshot: **633 slots**
- Current registry count: **620**
- Distribution cursor stuck at **620**, with slots **620–632 cleared**
- Repeated successful `DistributeEpoch` transactions supplied no remaining accounts and made no progress

We patched both tally and distribution to use the epoch snapshot limit and preserve one remaining-account entry per slot.

After deployment, our observer completed distribution in [this transaction](https://solscan.io/tx/22deCtmzXujr7DwDh9dxrfNiXPVraUddAY8HtDtnfU1UvZrcUP6FMN2j8L9rHyCP5QNhoFdnCykYBwuinqCUW4TV). Finalized state advanced **620 → 633**, and `rewardsDistributed` became **1**. The transaction performed no token transfers. Epoch **544** subsequently completed tallying and prescription with **50 observers**.

**Test scope:** this was an independent patch to SDK 4.1.2, using the fee payer as the filler account. I haven’t tested this PR’s branch or its read-only program-address filler.

**LLM disclosure:** An LLM coding assistant performed the investigation, wrote the patch and tests, and carried out verification and deployment under my direction. This comment was also drafted by the assistant.

vilenarios:
**Proven on-chain: this branch cleared a real D4 stall on staging.**

Staging epoch 820 hit the D4 shape naturally — crankers ran `finalize_gone` on 27 gateways mid-epoch, so the cursor stopped at **620/647** with slots 620–646 zeroed. The deployed crankers (pre-#726 SDK) sat there. I ran **this branch's `crankEpochStep()`** — unmodified, real signer, real sends — one step at a time:

```
epoch 820 before: dist_idx=620/647 distributed=0
step 1: action=distribute  tx 4J9o3RgSVU7kuEJB6eygrfTtPt5nsNPWG5D5wzKJTSttqXkYMTfcup21vzPBtHVy9AdWNcCe6m3W56SDpedZPGhm  -> cursor 638/647
step 2: action=distribute  tx 5mBSUAQSopoEPDjRtbCHJFZxQJbFZK6UA3eGiAndUsJN47a2Bd6ozkFh3rxnrYvNisgJ1H1s5EoydsG2csKsqj29  -> cursor 647/647 distributed=1
```

Verified from the landed transactions, not just the SDK's return value:

| | step 1 | step 2 |
|---|---|---|
| result | `err: null` | `err: null` |
| GAR-program fillers | 18 | 9 |
| GAR key writable in message | **no** | **no** |
| token balance changes / CPIs | none / 0 | none / 0 |

Both halves of the PR exercised for real: **D4** (range capped by `activeGatewayCount`, read-only fillers for zeroed slots) and **D4a** (cursor re-read and asserted after each batch — it passed because the cursor advanced; in an earlier capture-mode run where nothing was sent, it threw `did not advance: cursor still 620/647`, as designed).

For contrast, mainnet epoch 543 (same shape, 13 zeroed slots) sat stalled for ~3h20m while the deployed crankers sent `distribute_epoch` with zero remaining accounts — successful transactions that advanced nothing — until a gateway operator padded it by hand.

vilenarios:
@HarukaMa — thank you, this is a genuinely useful independent confirmation, and it lines up exactly.

I had decoded your transaction off-chain a few hours before your comment landed, while working out who cleared 543 (our crankers were still sitting at 620 supplying zero accounts). Chain-side it reads:

```
22deCtmz… payer 4KxtE7VN…  (registry slot 510)
  distribute ix: 23 accounts — 10 fixed + 13 remaining
  remaining: 13x 4KxtE7VN… (the fee payer)
  inner CPIs: 0    token balance changes: none    CU: 28394
  cursor 620 -> 633, rewards_distributed 0 -> 1
```

Worth noting for anyone comparing approaches: **the filler choice is cost-neutral.** Your fee-payer filler landed at 28,394 CU; this branch's read-only program-address filler simulated at 28,411 CU against the same state — 17 CU apart. Both work because the on-chain zeroed-slot skip runs *before* the `owner` and `is_writable` checks, and both reuse an account already in the message, so neither adds an account key against `MAX_TX_ACCOUNT_LOCKS`.

One portability detail if you ever switch fillers: with `@solana/kit` the program address must be attached **read-only**. Marking it writable fails at message-compile time with *"Program addresses may not be writable"* — it never reaches the network. The fee payer doesn't hit that because it's writable by necessity. That bit me on the first version of this branch.

## The half that may be missing from your fork — worth adding

Your description covers the padding (what this PR calls D4). This PR has a second part, D4a, and it is the safety-relevant one for a node running unattended:

After every batched tally/distribute, re-read the epoch and **throw if the cursor did not advance**. The failure mode is a *successful* transaction, so a stalled cursor is the only available signal — no try/catch and no error-rate alerting can see it.

Throwing (rather than logging and continuing) is deliberate: it stops the crank tick **before** `createEpoch`. That ordering is load-bearing. `weights_epoch` is a single shared field per gateway, re-stamped by whichever epoch is tallied most recently, so creating the next epoch and tallying it while the current one is undistributed **destroys the current epoch's weights**, and the epoch can then never be paid correctly. That is exactly how mainnet **540** was lost — it stalled, the next epoch was created and tallied, and its remaining rewards became unrecoverable. 543 survived its 3h20m stall only because every client in play happened to wait for distribution to finish before creating 544.

If your patch pads but doesn't assert progress, the next variant of a silent stall will sit unnoticed, and whichever client creates the following epoch first can kill it.

## Two heads-ups since you're on 4.1.2

- **Mainnet `ario-gar` was upgraded on 2026-09-15** (slot `447280441`) with ADR-0032 and ADR-0033. Distribution no longer aborts the whole batch on a mid-epoch joiner — it skips it, pays 0, and continues — and only the live epoch may be tallied. Two additive errors: `EpochWeightsClobbered` (6097) and `EpochNoLongerLive` (6098). Both mean an epoch needs a human, not a retry.
- **If your fork classifies Anchor error codes numerically, import them instead.** Anchor assigns `6000 + variant-index`, so a variant inserted anywhere but the end shifts every later code. Our own standalone cranker had drifted two positions and was suppressing four real errors as "already done" while reporting benign races as incidents. `@ar.io/solana-contracts` >= 1.3.0 exports the constants (`ARIO_GAR_ERROR__*`) generated from the shipped IDL.

This PR is green and waiting on review; the intent is that nobody needs to run a fork for this.

PR #733, opened 2026-09-16
perf(solana): batch getPrimaryNames processId enrichment

vilenarios:
## What

`getPrimaryNames()` scans the core program once, then issues **one `getAccountInfo` per name** to attach `processId` — O(n) sequential round trips for a single call:

```ts
const items: PrimaryName[] = [];
for (const { data } of accounts) {
  try {
    const pn = deserializePrimaryName(data);
    const rec = await this.getArNSRecord({ name: baseNameOf(pn.name) });  // <- one RPC each
    items.push({ ...pn, processId: rec.processId });
  } catch { /* skip */ }
}
```

This batches the enrichment into `getMultipleAccounts` through a new `getArnsRecordProcessIds` helper, using the same bounded pool as `getGatewayAccumulators`.

**250 names: 250 sequential reads → 3 concurrent ones.** Same records, same forgiveness for orphans.

## Why

On mainnet this is ~230 reads per call. Measured on a portal service that polls this method every 10 minutes, that single loop was **~81% of the application's entire RPC bill** — ~65,000 calls a day across two clusters, to attach one field.

It's visible in a packet capture as a flat block of byte-identical requests, since every ArNS record PDA read is the same length.

## Why batched reads rather than a registry scan

A whole-program scan of the ArNS registry would also collapse this to one call, but it reads every record to use a few hundred. The batched path reads exactly the records needed — **strictly less data as well as fewer calls**. This is the break-even `getArNSRecordsByAntMints` already documents:

> Cheaper than scanning the whole registry as soon as the caller has fewer mints than the registry has records.

There's also a short circuit: no primary names means no batch read at all.

## Also: a latent correctness bug in the same join

The base name was resolved by an inline `split('_')` with a `parts.length === 2` check. That returns the **undername** rather than the base for any name carrying two or more underscores:

| name | inline rule | contract `splitn(2, '_')` |
|---|---|---|
| `alice` | `alice` | `alice` |
| `sub_alice` | `alice` | `alice` |
| `a_b_c` | **`a`** | **`b_c`** |

So `a_b_c` looked up the wrong record, silently failed, and was dropped from the results entirely.

It now uses `splitPrimaryName` — the shared helper that already encodes the contract rule and is used by `resolveArNSName` and the primary-name write path. **No name on mainnet or devnet carries two underscores today**, so this corrects a latent bug rather than changing any current output.

Happy to split this into its own PR if you'd rather keep the perf change isolated — it's bundled because it's the same two lines the rewrite touches, and shipping a knowingly-wrong join key seemed worse.

## Tests

**572 total, 571 pass, 0 fail** (baseline on `main`: 568 pass, 0 fail). Three added, following the existing `countingRpc` / stub-rpc pattern in `io-readable.test.ts`:

- the contract split rule — `a_b_c` must resolve through `b_c`, and an orphan is still skipped
- the batching counts — 250 names produce 3 `getMultipleAccounts`, 250 accounts, 1 program scan, and **zero `getAccountInfo`**
- the empty-scan short circuit — no batch read when there is nothing to enrich

`tsc -p tsconfig.json --noEmit` clean. `biome check --unsafe` on both touched files reports the same 1 pre-existing warning as before the change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved primary-name resolution for names containing multiple underscores.
  - Primary-name lookups now gracefully skip invalid or undecodable records instead of failing.
  - Empty name requests return results without making unnecessary network calls.

- **Performance**
  - Process ID resolution now uses batched account lookups with bounded concurrency, reducing the number of network requests and improving responsiveness.


vilenarios:
Two follow-ups after a closer pass, both now pushed.

**1. One deliberate behavioural difference worth a reviewer's eye.** In the old loop, the per-name `getArNSRecord` sat *inside* the `try`, so **any** error for a single name — including a transient RPC failure — was swallowed and that name was silently dropped. The batched path only catches deserialization; a fetch error that survives `withRetry` now propagates out of `getPrimaryNames()`.

I think that is the right direction (an RPC blip previously produced a quietly incomplete result, which is the worse failure), and it matches how `getGatewayAccumulators` already behaves — it catches deserialize errors but lets `withRetry(fetchEncodedAccounts(...))` propagate. Flagging it explicitly since it is not a pure performance change. Happy to wrap it if you would rather preserve the old swallow-everything semantics.

**2. Coverage.** Codecov flagged 7 uncovered lines; only one was actually in the new code (the rest belong to the pre-existing `getPrimaryNameRequests` directly above). The gaps were in the branches deciding whether a bad account takes the whole call down, so they are now pinned by three tests — an undecodable primary-name account, an undecodable ArNS record, and `getArnsRecordProcessIds([])` making no round trip. **New code is at 100% statement and branch coverage** (lines 2927-3037).

Note the project-coverage status is red against an 80% target, but that is pre-existing and repo-wide: base is 61.07% and this PR takes it **up** to 61.33%.

Also checked and clear: `alpha` is 2 commits ahead of `main` and neither touches `io-readable.ts`, so there is no conflict on the base.

vilenarios:
@coderabbitai review

vilenarios:
Good catch, @coderabbitai — that was a real inconsistency and it was mine. Fixed in `ca6634d`.

`getPrimaryName` (singular) carried its own inline copy of the same rule, so after the first commit the two public readers disagreed about which `ArnsRecord` the same name resolves through: plural resolved `a_b_c` via `b_c`, singular still via `a`. That is worse than the original bug, because inconsistent beats merely wrong only in the wrong direction.

Both now go through `splitPrimaryName`. I grepped for other copies — `parts.length === 2 ? parts[1] : parts[0]` and every `split('_')` in `src/` — and that was the last one; there are no others.

Added a test driving the singular by-address path and asserting `a_b_c` resolves through `b_c`, so the two readers cannot drift apart again.

576 tests, 575 pass, 0 fail (baseline 568 pass, 0 fail). Typecheck clean, biome back to the same 1 pre-existing warning.

@coderabbitai review

vilenarios:
@coderabbitai review

Two things have landed since your last pass (which was on `e4abe90`):

- `ca6634d` — the fix for the `getPrimaryName` (singular) inconsistency you flagged. Both readers now go through `splitPrimaryName`; I grepped and that was the last inline copy of the rule in `src/`.
- `b10e9eb` — `alpha` merged into the branch.

Please take another look at the head commit.

